PR #38, opened 2017-09-09
Arrow down & Enter; FlyTo address result

### LINKED ISSUE #18: Search results can't arrow down / press enter
I can't arrow down on the results or hit enter for what I've entered

allthesignals:
De-abstract some things (wrong direction); introduces new methods in map-search, more event handlers for keyboard strokes. main-map now has access to a general routeTo action.

Would like @andycochran's design input on arrow down highlighting. Also, please test numerous addresses to check for strange behavior.

Closes #18

### DIFF
--- a/app/components/main-map.js
+++ b/app/components/main-map.js
@@ -234,6 +234,8 @@ export default Ember.Component.extend({
 
   actions: {
     handleMapLoad(map) {
+      const mainMap = this.get('mainMap');
+      mainMap.set('mapInstance', map);
       map.addControl(new mapboxgl.NavigationControl(), 'top-left');
       map.moveLayer('building');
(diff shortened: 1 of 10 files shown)


chriswhong:
Added a 300 pixel padding to the `fitbounds()`, conditional on whether `mainMap.selected` exists.

chriswhong:
When selecting an "address" result, there should be a route change.  When I selected an address after having a lot already selected, the map navigates to the selected address, but the route and the data in the UI reflect the previous lot.

allthesignals:
@chriswhong agreed, but this PR is really about the arrow key up and down and pressing enter issue - should we merge since it solves that? I'm going to open a separate issue for your comment so we can work on it separately.

allthesignals:
Merging this since I'm not hearing any more feedback on the arrow and enter functionality.

PR #39, opened 2017-09-10
fix partial lot highlighting

chriswhong:
This PR fixes an issue where lots that are highlighted on mouseover would only partially render due to proximity to a tile edge.  

Instead of only taking the [0]th feature when using `queryrenderedfeatures`, all matching features are added to the map source for highlighting (the only situation where queryrenderedfeatures should return more than one feature is in this tile edge scenario)

Additionally, this PR also "unhighlights" the last highlighted lot when the user mouses out of the lots layer.

### DIFF
--- a/tests/unit/utils/search-pluto-lots-test.js
+++ b/tests/unit/utils/search-pluto-lots-test.js
@@ -5,6 +5,6 @@ module('Unit | Utility | search pluto lots');
 
 // Replace this with your real tests.
 test('it works', function(assert) {
-  let result = searchPlutoLots();
+  let result = searchPlutoLots('120 broadway');
   assert.ok(result);
 });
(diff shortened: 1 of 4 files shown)


tests/unit/utils/search-pluto-lots-test.js
```diff
@@ -5,6 +5,6 @@ module('Unit | Utility | search pluto lots');
 
 // Replace this with your real tests.
 test('it works', function(assert) {
-  let result = searchPlutoLots();
+  let result = searchPlutoLots('120 broadway');
```
allthesignals: This works fine, but in #38 I simply set a default value for the utility (which is a blank string). Thinking out loud here, but this is where [TypeScript](https://en.wikipedia.org/wiki/TypeScript) comes in handy by giving our JS strongly typed classes.

PR #61, opened 2017-09-14
Layer Palette Stlyes

andycochran:
This PR adds styles for the `.layers-list` and `.layer-group`, adding a foundation switch to indicate the layer state.


andycochran:
@allthesignals Yay! Tests are passing. 🎉

Issue #45, opened 2017-09-11
Complex Search Prototype


chriswhong:
User should be able to see many "types" of typeahead results in the search pane.

- [x] Lots - PLUTO
- [x] Mapzen search
- [x] Zoning Map Amendments - Route to individual Zoning Map Amendment

chriswhong:
Planning to implement a typeahead search web API that will query several sources simultaneously and return a results object ready for display in the app with the following result types:

- Lots
- Zoning Actions
- Block

andycochran:
I want to hit `ESC` to clear the search.

chriswhong:
Closed by #62

PR #62, opened 2017-09-15
45 composite search

chriswhong:
Updates search component:
- Consumes `https://github.com/NYCPlanning/labs-zola-search-api` which allows for searching bbls, addresses, and ulurp numbers.

- Allows user to hit escape to clear the results list

- Added fa-icon component and reformatted results


chriswhong:
Closes #45

PR #109, opened 2017-09-19
93 layer controls

allthesignals:
This pull request adds several new components as part of a "layer control" API:
`layer-group`,  `layer-control-timeline`, `layer-multi-select-control`, `layer-checkbox`, `group-checkbox`, `cascading-checkbox`

See examples for each:
# Layer Group
`layer-group` sets up map layers in mapbox-gl based on a given layer definition file. It is a sub-component of `mabox-gl` and must be called as `map.layer-group`.
Properties:
 - `config` (required) - this is accepts a `layer-group` configuration, which is an extension of mapboxGl's layer source definition. You can generate these with `ember g layer-group {layer-group-name}`
- `qps` (optional) - query params - this should be the controller scope so that the `layer-group` can bind its state. Optional.

Example:
```
{{#mapbox-gl as |map|}}
  {{map.layer-group config=pluto qps=qps}}
{{/mapbox-gl}}
```

## Block Context
`layer-group` yields some of its internals through [contextual components](https://guides.emberjs.com/v2.15.0/components/wrapping-content-in-a-component/#toc_sharing-component-data-with-its-wrapped-content). It also includes subcomponents.

### Properties / Actions
`visible` - [property] whether layer is visible or not (boolean, true / false)
`update` - [action] `layer-group`'s internal action for updating SQL. It's signature is:
`updateSql(method = 'string', column = 'string', value = [])`, where `method` is any of the named methods (`buildRangeSQL`, `buildMultiSelectSQL`, for now). 
`toggleVisibility` - action that may be passed to toggle layer visibility. 
`isPending` - boolean (true/false) layer-specific loading state (only the state of the initial handshake, not tiles!)

### Sub-Components
`timeline-control` - (details below)
`multi-select-control` - (details below)

### Example:
```
{{#map.layer-group 
  config=zoningMapAmendments 
  tagName='li' 
  classNames='layer-group clearfix'
  qps=qps as |group|}}

  {{#if group.visible}}
    {{group.timeline-control
        column='effective'
        query-param='zma-effective'}}
  {{/if}}
{{/map.layer-group}}
```

# Timeline Control
Allows for a range slider with two handles that filter a carto-type layer by a given `date` column. 

`column` = SQL column with the date value
`query-param` = control's named query parameter 

Example:
```
{{#map.layer-group 
  config=zoningMapAmendments 
  qps=qps as |group|}}
    {{group.timeline-control
        column='effective'
        query-param='zma-effective'}}
{{/map.layer-group}}
```

# Multi Select Control
Allows for carto-type layers to be filtered by a list of values. Requires configured subcomponents. Exposes a `layer-checkbox`:
```
{{#group.multi-select-control 
column='overlay' as |multiSelect|}}
    {{multiSelect.checkbox 
        value='C1-4'}} 
      C1-4 
{{/group.multi-select-control}}
```
Requires a `column` property that specifies the SQL column used to filter. 

# Group Checkbox
This is the explicit implementation of cascading checkboxes. The other component, cascading-checkboxes, implements this in a more expressive way, but it is hacky and doesn't support deeply nested logic. We should use this explicit group-checkbox until cascading-checkbox can be refactored and cleaned up. 

In the example below, each layer-checkbox is given an explicit value for reference. These values do not have to be defined upfront, they just can't collide with anything else in the parent scope. 

group-checkbox requires `refs`, `values`, and a `scope`. `refs` should be an array of string values naming the variables used for the group logic. `values` should be an array of the values themselves, passed as variables. The `scope` is the object on which values `get` and `set`. 
```
<ul>      
  <li>
    {{group-checkbox
      refs=(array 'c14' 'c24' 'c22' 'c13' 'c23')
      values=(array c14 c24 c22 c13 c23)
      scope=this}}
  </li>
  <li>
    {{multiSelect.checkbox 
        value='C1-4'
        checked=c14}} C1-4 
  </li>
  <li>
    {{multiSelect.checkbox 
        value='C2-4'
        checked=c24}} C2-4
  </li>
  <li>
    {{multiSelect.checkbox 
        value='C2-2'
        checked=c22}} C2-2
  </li>
  <li>
    {{multiSelect.checkbox 
        value='C1-3'
        checked=c13}} C1-3
  </li>
  <li>
    {{multiSelect.checkbox 
        value='C2-3'
        checked=c23}} C2-3
  </li>
</ul>
```

# Get Unique Options For 
(Helper) Gets unique list options for a given column in a given `sql` query. 
```
{{get-unique-options-for 'primaryzone' zoningDistricts.sql}}
```

# Extract Layer Stops For
(Helper) This grabs the MapboxGL style "stops" from a given `layer-group` and `layer` id. 

```
{{extract-layer-stops-for 'layerId' layerConfigObject}}
```


allthesignals:
I need to fix tests but please take a look

chriswhong:
All looks good to me, looking forward to start using these.

PR #112, opened 2017-09-19
closable content area

andycochran:
This PR makes the `.content-area` element "closable" by moving the welcome message to the about route, adding a `.content-is-closed` class to the `body` on the index route, and adding styles to hide the `.content-area` element.


andycochran:
@allthesignals This still needs work… 

The `.mapboxgl-canvas` element remains the width of its container when the app is first loaded. So when you open/close the content area, the map can be too wide/narrow. 

Two ways to reproduce this: 
- Refresh the app with a about page route. Click the close button in the content area to go to the index route. The map does not fill its allotted space. 
- Refresh the app with the index route. Load a tax lot. The map thinks it's wider than it actually is, so the lot isn't centered appropriately.

^ In either case, resizing the window triggers a map resize and fixes the map's layout. Can we fire a [resize](https://www.mapbox.com/mapbox-gl-js/api/#map#resize) when the route changes?

allthesignals:
@andycochran Great - let me know if I can help with these tests!

andycochran:
@allthesignals Test are passing now. But I could use help with the map resize described above.

andycochran:
`map.invalidateSize();​`?

allthesignals:
@andycochran this looks great - I'm just a little concerned about `routes/index.js` and the jQuery usage, mainly because it can get confusing later on... is there a way we can insert the `.content-area` div inside the lot and zma view? Let's put in a pin in it for later if not but want to make sure.

Thanks.

allthesignals:
Thanks, @andycochran - I'll try to re-arrange it, sorry about that. jQuery is fine but only really in components only because of the async way things get rendered.

allthesignals:
Hey @andycochran can you take a look at what I pushed? I used an auto size cell (http://foundation.zurb.com/sites/docs/xy-grid.html#auto-sizing). It seems to be working okay, but it doesn't adjust the fit founds, and some of the responsive views don't look right. Let me know if this approach is possible... if we need to include classes dynamically, let's look into an ember addon. Thanks

andycochran:
@allthesignals I've fixed the responsive layouts. Passing off to you to add the `.resize()` on route change.

PR #120, opened 2017-09-21
91 subway layer

chriswhong:
Refactors `layer-group` to take a SQL array for type `carto`, allowing for creation of vector tiles with multiple inner layers.

Adds subway lines, stops, and labels copied from capital planning platform: https://github.com/NYCPlanning/capital-planning-platform/blob/master/app/janelayers/transportation/config.js


allthesignals:
Sweet! Tests fail because there are two console errors: `TypeError: Cannot read property 'map' of undefined`.

I think this happened in b8b3ec7.

PR #127, opened 2017-09-21
checkbox labels

andycochran:
This PR… 
- makes the Layer Palette wider on large screens
- wraps each checkbox with a `label` element, which allows the whole thing to be clickable
- adds a new `.layer-menu-item-controls` class and styles for labels it contains


chriswhong:
Looks great!

PR #128, opened 2017-09-21
99 tooltips

allthesignals:
This adds very basic tooltip support. See:
```
 {{#hover-tooltip as |tooltip|}}
    {{#if (eq tooltip.layer 'pluto-fill')}}
      {{tooltip.layer}}: {{tooltip.feature.properties.bbl}}
    {{/if}}
  {{/hover-tooltip}}
```


chriswhong:
So take a look at 1591013605fca02d820c1c99a30b3ea7ba9cbb59, I did some refactoring:

The `mousemove` event kicks everything off, of course.  When the mouse moves, it runs `highligher()` in the `map-mousemove` service.  We had layers hard-coded in here, but I added a computed property in `registered-layer` to get what I am calling "highlightableAndVisibleLayerIds".

Basically, when we run `queryrenderedfeatures()`, we only want to query layers that are highlightable AND visible.

`highlightable` is a new config property, which goes on the layer level, not the layergroup level.  Basically, any layer that should get a tooltip and a highlight should have `highlightable =true` in its config.

Next, once queryrenderedFeatures gets results back, it just takes the first one, which will be the top-most layer.

It then takes this feature and assigns it to `highlightedLotFeatures`, but here is where things get different, we use this highlighted feature to drive BOTH the tooltip and the creation of the highlight layer on the map!

`hover-tooltip` has logic to only show {{text}} when the highlightedfeature exists.  But how does it know what to show?  I added another config property called `tooltiptemplate` which gets rendered with mustache.  The `tooltipTemplate` for pluto looks like: `tooltipTemplate: '{{address}}',`

So, in summary, each time the mouse enters a `highlightable` and visible layer, `map-mousemove`grabs the feature AND renders the text for the tooltip.

allthesignals:
AWESOME.

PR #130, opened 2017-09-22
new search layout

andycochran:
This PR revises the search layout. 

It's currently ridiculously wide: 

image

Now that the layer palette is wider, it can match that width: 

I've also added a search icon when there are no terms:


andycochran:
@chriswhong Test are passing now. Merge?

PR #134, opened 2017-09-22
Bind QPs, reset QPs on transition out of Lot

### LINKED ISSUE #80: Ability to reset map view state (not necessarily filters)

allthesignals:
Closes #80 - resets QPs when transition out of lot view


chriswhong:
I think clicking "ZoLa" should reset the app to the default state regardless of what route the user is on.

chriswhong:
Looks great!

PR #136, opened 2017-09-22
94 admin boundaries

chriswhong:
Adds many admin boundary layers with common styling.
- Boroughs
- Community Districts
- Council Districts
- State Senate Districts
- State Assembly Districts
- NTAs


allthesignals:
Yay! Only thing: admin-boundaries-styles should be a utility rather than a helper - it should also be more generic than it is currently or we need a new generator for sharing definitions.

Issue #111, opened 2017-09-19
Layer ordering should be based on order in which they appear in template


allthesignals:
Layer's currently re-draw on top of existing layers when any computed property is updated

PR #153, opened 2017-09-24
Add fixed layer order

chriswhong:
This PR adds a fixed order to the layergroups in the app, allowing for toggling layergroups on and off and ensuring that they end up in the same place relative to other layergroups.

Layer order is defined by the order that layergroups are listed in `main-map.hbs`.  (The first layergroup in the template will be the top-most on the map, with the aerial rasters being last)

To accomplish this, we added a computed `before` property in `layer-group.js` which walks through all of the layergroups that are higher than the one being rendered.  If the layergroup's bottom-most layer exists in the map style, it uses that layer as the `before` property.  If not, it tried the next highest layergroup.  

If none of the higher layergroups has any layers on the map, it will default to using `waterway-label` as the `before` value.

*Things to consider*
- Highlighted and Selected layers are not rendered using `layer-group`, so they live outside of this framework.  Right now they are rendered `before` label-waterway, and after all of the other layergroups, so they will always render "on top"

- However, I think it would be a better-looking solution to make a highlight layer take the same position as the layer it is highlighting.  We might be able to use `moveLayer()` (which takes a `before` argument just like `addLayer()`) whenever the highlight feature is updated.

### DIFF
--- a/app/components/layer-group.js
+++ b/app/components/layer-group.js
@@ -69,7 +69,7 @@ export default Ember.Component.extend(ParentMixin, ChildMixin, {
     const position = allLayerGroups.map(layerGroup => layerGroup.config.id).indexOf(this.get('config.id'));
 
     // walk all layergroups that should be displayed above this one
-    for (let i = position + 1; i < allLayerGroups.length; i += 1) {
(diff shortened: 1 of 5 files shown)


allthesignals:
Closes #111

Issue #162, opened 2017-09-25
NTA layer shows bad labels


andycochran:
e.g. Manhattan labels show in Queens/Brooklyn on the waterfront.

andycochran:
Manhattan labels in Queens: 

![image](https://user-images.githubusercontent.com/409279/30818608-02b84cec-a1ea-11e7-9a2b-16f91dd054fa.png)

Nix "parks cem etc" labels

![image](https://user-images.githubusercontent.com/409279/30818640-14aafa76-a1ea-11e7-886b-dbe15bb499c2.png)

chriswhong:
Two fixes here:
- [ ] Add a centroids-only point layer to the vector tiles, and use that for labeling
- [ ] Remove the non-nta areas from the query (park-cemetery-etc features should be excluded)

chriswhong:
Closed by #169

PR #169, opened 2017-09-26
add centroids data, remove non-ntas

chriswhong:
Adds a centroids layer to the NTA vector tile for use in label placement,
removes non-nta features from query (parks, cemeteries, etc)


chriswhong:
Closes #162

PR #172, opened 2017-09-26
159 address point

### LINKED ISSUE #159: Design/UX issue: solution for search with address problem
Search by address can be confusing because when you search and get a Mapzen address, you "don't get information about this property". This leads to possible confusion for the user. 

One solution would be to plot a point on the map indicating something has happened.

allthesignals:
Closes #159 

Plots point on map when searching address. 

This also attempts to address fitBounds & map re-sizing across route transitions...

Current working theory is many competing events are grabbing state from different places. Calling `resize()` won't always fix the issue, but then it depends on the anticipated amount of time it will take for a `fitBounds` method to trigger `boxzoomend` or `moveend`. What's probably happening is that `#resize` is called but only after a route transition and _after_ fitBounds has been fired. My guess is that fitBounds is using it's old context and doesn't adjust it's bounding context mid-flight. 

Calling it afterward depends on inserting a number of awkward `.once('moveend, fn)`... this seemed to work but was behaving oddly.

Instead it is doing this in the component on a `div`-sensitive resize event... (as opposed to window). We should keep investigating this issue as resize happens only 95% of the time (because of variable `fitBounds` transition animation lengths).


andycochran:
@allthesignals,  

I've fixed the issue with opacity/focus (`.has-results` was being used to toggle the opacity of the results on focus/blur of search input so they were still there/clickable, covering the filters). I made a separate `.focused` class that hides the search results altogether. 

Tho there are still two things we might want to tidy up (maybe in separate issues?): 

- When a result is selected (e.g. "200 Broadway, Manhattan, New York, NY, USA"), the results are hidden but the input only shows the text as typed (e.g. "200 Broa"). Can we update the input to match what was selected? 
- When you click a result, `handleFocusOut` is triggered. But when you arrow down and hit enter it's not. You're still focussed and the results remain visible. Should we programmatically blur the input?

Issue #145, opened 2017-09-22
Link to current digital tax map from lot view


allthesignals:
Current ZoLa links to digital tax maps. Replicate this feature.

chriswhong:
Closed by #174

PR #174, opened 2017-09-27
add DTM, zonemap, and historical zone map links

chriswhong:
Adds link to selected lot view on the official city DTM web map.

Also added links to the zoning map PDF and the historical zoning map PDF


chriswhong:
Closes #145

Issue #144, opened 2017-09-22
"What are 'transit zones'?"


allthesignals:
Might need description of various layers like "Transit Zones" and "FRESH Zones", or a link to DCP site...

chriswhong:
Implemented layergroup title tooltips in #175, added #176 to add more title tooltips in next sprint

PR #175, opened 2017-09-27
144 layer tooltips

chriswhong:
This PR adds a `titleTooltip` property to `layer-group` configs.  If it exists, the app will render a tooltip next to the layer's title in the layer palette.
screen shot 2017-09-27 at 12 33 55 am


chriswhong:
Closes #144

Issue #161, opened 2017-09-25
Add subway entrances to transit layer.


andycochran:
Add subway station entrances, make them appear when zoomed in (and make station circles disappear)

chriswhong:
It would also be nice to add the vectors of the station outlines, but I am not sure these are #opendata.
[Uploading Google_Maps.png...]

chriswhong:
Closed by #179, Subway entrances are indicated with markers, can re-open if we discover a source for station footprints data.

PR #179, opened 2017-09-27
add subway entrances data, add icons and labels

chriswhong:
This PR adds subway entrances data, icons, and labels to the map, they fade in after zoom level 15.5.

labszola


chriswhong:
Closes #161

PR #187, opened 2017-09-27
79 center zoom hash

chriswhong:
This PR adds the mapboxGL map hash, allowing us to persist map views on page reload.

It also changes the fitbounds behavior
- Transitioning from sub-routes to index no longer zooms out
- Highlighted features are cleared when transitioning back to index 
- Fitbounds adds an offset option so the fit is correct even though the map is obscured behind the details pane.

TODO: Add logic to `fitBoundsOptions` to make sure it works on mobile


andycochran:
The © info and map scale are hidden behind the content on > large screens

andycochran:
Seems like the map pan/zooms on route change on very large screens but not on smaller screens.

andycochran:
This needs to account for the `large` and `xxlarge` CSS media queries: 

```scss
$breakpoints: (
  small: 0,
  medium: 640px,
  large: 1024px,
  xlarge: 1200px,
  xxlarge: 1440px,
);
```

app/components/main-map.js
```diff
@@ -43,8 +43,16 @@ export default Ember.Component.extend({
       const height = el.height();
       const width = el.width();
 
+      const fullWidth = $('.navigation-area').width();
```
allthesignals: @chriswhong is this the only div where we can get correct information about the width of the screen?

andycochran:
To keep the media queries in sync with the JS consider using this method to have the CSS tell the JavaScript which media query is being used: http://zerosixthree.se/detecting-media-queries-with-javascript/

chriswhong:
Will tackle tomorrow, not ready to merge yet

app/components/main-map.js
```diff
@@ -43,8 +43,16 @@ export default Ember.Component.extend({
...
+      // width of content area on large screens is 5/12 of full
+      const contentWidth = (fullWidth / 12) * 5;
+      const offset = -((width - contentWidth) / 2);
+
       const padding = Math.min(height, width) / 2.5;
-      return { padding: selected && (type !== 'zoning-district') ? padding : 0 };
+      return {
+        padding: selected && (type !== 'zoning-district') ? padding : 0,
+        offset: [offset, 0],
+      };
```
allthesignals: @chriswhong what should this computed be doing in terms of the view? Having trouble figuring out how to test / verify. Thanks!

allthesignals: ^^ I have some ideas on how to approach this differently

app/components/main-map.js
```diff
@@ -63,6 +71,11 @@ export default Ember.Component.extend({
   },
   highlightedLotLayer,
 
+  @computed('mainMap.shouldFitBounds')
+  shouldFitBounds(shouldFitBounds) {
+    return shouldFitBounds;
+  },
```
allthesignals: I'll refactor this to a computed alias (shrink 4 lines of code into 1)

app/routes/index.js
```diff
@@ -0,0 +1,14 @@
...
+const { service } = Ember.inject;
+
+export default Ember.Route.extend({
+  mainMap: service(),
+
+  actions: {
+    didTransition() {
+      this.get('mainMap').set('selected', null);
+      this.get('mainMap').set('shouldFitBounds', false);
+    },
```
allthesignals: I'll take care of this but if you're setting on a property you don't need to get it, you can just write:
`this.set('mainMap.selected',null);`

Alternatively, if there are multiple properties, you can just say:
```
      this.get('mainMap')
        .setProperties({
          selected: null,
          shouldFitBounds: false,
        });
```

app/routes/lot.js
```diff
@@ -14,5 +14,6 @@ export default Ember.Route.extend({
   afterModel(model) {
     const mainMap = this.get('mainMap');
     mainMap.set('selected', model);
+    mainMap.set('shouldFitBounds', true);
```
allthesignals: Same as above, I'll take care of it tonight:
```
      this.get('mainMap')
        .setProperties({
          selected: null,
          shouldFitBounds: false,
        });
```

app/routes/zma.js
```diff
@@ -15,9 +15,8 @@ export default Ember.Route.extend({
   },
 
   actions: {
-    willTransition() {
-      const application = this.controllerFor('application');
-      application.resetQueryParams();
+    didTransition() {
+      this.get('mainMap').set('shouldFitBounds', true);
```
allthesignals: `this.set('mainMap.shouldFitBounds', true);`

app/routes/zoning-district.js
```diff
@@ -13,4 +13,10 @@ export default Ember.Route.extend({
     const mainMap = this.get('mainMap');
     mainMap.set('selected', model);
   },
+
+  actions: {
+    didTransition() {
+      this.get('mainMap').set('shouldFitBounds', true);
```
allthesignals: ditto

app/components/main-map.js
allthesignals: Basically, `shouldFitBounds: alias('mainMap.shouldFitBounds')`

app/components/main-map.js
chriswhong: No, it's just the closest one.  Just need something full width so we can calculate the percentage that the details pane will take up.

app/components/main-map.js
chriswhong: This offsets the fitbounds, I am not sure how you would test the output, you would need to be able to read the centerpoint of the map vs the center of the highlighted polygon somehow.

app/routes/index.js
chriswhong: good to know!

andycochran:
I pulled down the latest on this and got a weird merge conflict. I fixed the `_l-default.scss` file, but GitHub says my commit touched other files. Please confirm I didn't break anything. 😬

app/components/main-map.js
allthesignals: Got it - so I think window width will do.

app/components/main-map.js
allthesignals: Ah, I just mean manual testing. Why does zoning-district get padding but not others? Thanks!

allthesignals:
@chriswhong @andycochran 
I finally successfully solved the resize issue which is described in the commit 
(4284b18). However, it seems like the strategy of this PR was to solve it by working around it, and making the map absolutely positioned so that `.content-area` would float over it. That means that even though the original behavior works (resize map when a new div is inserted), the map center still jumps to adjust for the container's new width, moving layers and contently along with it. 

There is a long discussion thread on this related issue here: https://github.com/mapbox/mapbox-gl-js/issues/1740. Basically, MapboxGL does not yet provide a way to globally offset the map's internal concept of a "center", although it does provide an "around" option (https://www.mapbox.com/mapbox-gl-js/api/#cameraoptions) for a select few panning/zoom methods. 

I think this will have to be the working solution for now but we'll need to find a more organized way to deal with the layout widths to calculate offsets since that will cause issues for mobile and other views.

PR #200, opened 2017-09-29
make zoning fill fade out between zoom 15 and 16

chriswhong:
Fade zoning colors out as tax lots fade in, like:

sep-28-2017 22-16-20


allthesignals:
Gorgeous!

PR #216, opened 2017-09-30
Cw map improvements 2

chriswhong:
This PR refactors `layer-group` configs so that `sources` are decoupled from `layers`.  Under the status quo, a layer's config did not explicitly define a source, the layer's source was its parent (layers were rendered in handlebars as `{{source.layer}}`)

Most of the changes are modifications to layer-group configs. 

# Changes:

## Carto Handshake
`carto.getVectorTileTemplate()` now expects an array of `source-layer` objects (`{id, sql}`).

## Sources  
- `app/sources` includes source configs, and the entire app's map data is now represented in 8 vector tile sources and 2 raster tile sources.  
- Sources include a type property of 'cartovector' or 'raster'.  cartovector sources become `vector` sources once the tile initialization is complete. 
- cartovector sources include a `source-layers` array, each source-layer is an object with `id` and `sql`.  The `id` allows us to give meaningful names to the source-layers within a carto vector tile instead of layer0, layer1, etc...
- Some logic was added to make sure that the map's layers don't render until all asynchronous sources are ready.  non-async sources are added to the map immediately.

A cartovector source looks like this: 

```
export default {
  id: 'transportation',
  type: 'cartovector',
  'source-layers': [
    {
      id: 'subway-routes',
      sql: 'SELECT the_geom_webmercator, rt_symbol FROM support_trans_mta_subway_routes',
    },
    {
      id: 'subway-stops',
      sql: 'SELECT the_geom_webmercator, name FROM support_trans_mta_subway_stops',
    },
    {
      id: 'subway-entrances',
      sql: 'SELECT the_geom_webmercator FROM support_trans_mta_subway_entrances',
    },
  ],
};

```

## Layer Groups
- Layergroups are literally groups of layers that only reference `source`, but don't contain any config for sources. 
- Most of the`layer-group` configs are the same, except layers now reference pretty names for their `source-layer`

A `layer-group` config for a vector source looks like this:
```
export default {
  id: 'historic-districts',
  title: 'Historic Districts',
  visible: false,
  layers: [
    {
      layer: {
        id: 'historicdistricts-line',
        type: 'line',
        source: 'landmark-historic',
        'source-layer': 'historic-districts',
        paint: {
          'line-width': 2,
          'line-color': 'orange',
        },
      },
    },
  ],
};

```

TODO
- update generator for layer-groups 
- add a generator for source


app/components/layer-group.js
```diff
@@ -134,37 +100,49 @@ export default Ember.Component.extend(ParentMixin, ChildMixin, {
...
   },
 
   actions: {
     toggleVisibility() {
       this.toggleProperty('visible');
     },
-    updateSql(method, column, value) {
-      const sql = this[method](column, value);
-      this.set('sql', [sql]);
+    updateSql(method, sourceId, column, value) {
```
allthesignals: We could get into trouble here with race conditions but this is not usually ever a long-running request so I'm not concerned.

allthesignals: I'm going to open an issue on ember-mapbox-gl, I think this is a use case that can nudge the addon in a new way.

allthesignals:
I made some changes, mostly moving the carto handleshake promises into the route (where we get promise resolution for free).

Issue #173, opened 2017-09-27
User sees PLUTO data in an established hierarchy in the lot view


chriswhong:
Which fields should we show?  How should they be organized?  What conditional notices should we show (waterfront lot, etc)?

allthesignals:
- Link out to zoning text

PR #230, opened 2017-10-03
173 pluto layout

andycochran:
This PR tweaks the layout and styles on the Lot view. Supplements #173, which is already closed but could've been prettier.


chriswhong:
These admin boundaries colors are 🔥

PR #231, opened 2017-10-03
141 zoning fitbounds

chriswhong:
This PR:

- Adds logic to leave the map alone when routing to a zoning district via map click
- Make the map fitbounds for the zoning district when routing from a search result
- Adds a "Fit map to all XX districts" button, so the user can see all if they choose.


andycochran:
^ This button behaves just like a search. Is that the desired behavior? (I think yes.)

Issue #196, opened 2017-09-28
Highlighted feature layer should appear just above the source layer


chriswhong:
The mapboxgl layer that we use for highlighted features is added to the map once, and its data is constantly changed as the highlighted feature changes.

This means the layer being highlighted may have others on top of it, but the highlighted feature appears on top.

We should explore using `layer.moveTo()` to figure out where the highlighted layer should go in the layer stack.

PR #234, opened 2017-10-04
move highlight layer on mouseover

chriswhong:
This PR changes the ordering position of the `highlighted-lot` layer during the mousemovehandler, causing it to be rendered directly above the layer it is highlighting.  (It used to always be "on top", which shaded layers that were rendered above the feature of interest.


chriswhong:
Closes #196

PR #251, opened 2017-10-05
add layer clickable, handle clicks dynamically

chriswhong:
This PR adds a `clickable` boolean to layers within layer-groups, and is used to make sure that the layer the user clicked on is indeed clickable.  No action is taken if it is not.

This is accomplished with the following logic:
- On mousemove, store the top-most visible layer id as `highlightedLayer` in the `mapMouseover` service
- On click, `queryRenderedFeatures` only for those layers that are `visible` AND `clickable`
- If the top-most clicked feature's layer id is the same as the mapMouseover.highlightedLayer, handle the click, otherwise, do nothing.


andycochran:
@chriswhong Can this also control if hovering a feature changes the cursor to `pointer` or `grab`?

chriswhong:
Yes, I think so... I assume you want it to only be a pointer if it is clickable?

PR #256, opened 2017-10-05
176 title tooltips

chriswhong:
First pass at titleTooltips for applicable layers.

Excluded layers: all admin boundaries, some subways.


andycochran:
If the tooltip wraps to the next line, the tooltip is positioned off the page. 



Does `ember-tooltip` have the ability to responsively position top/right/bottom/left?

andycochran:
Other than the above, this LGTM. I'll merge and open a separate issue for addressing the positioning.

PR #310, opened 2017-10-11
296 measuring tool

### LINKED ISSUE #296: Add measuring tool

allthesignals:
Closes #296 - needs some styling attention from @andycochran.


andycochran:
It might not be clear that you have to click the last point of a line to stop drawing that line. Or that you have to click the first point to close a polygon. Do we need instructions that are visible while you're drawing?

andycochran:
Pushed some style changes.

allthesignals:
"Need to not change routes while drawing." This is turning into a rabbit hole. I'll keep digging? MapboxGL has no concept of MapboxGL Draw - the latter is a child of the former. Since route changes are top-level, and occur through a delegated controller action, there's no shared state between the main-map's Draw state and the application controller, unless through a service. The problem with the service is there's no straightforward MapboxGL Draw even that tells when drawing has begin and ended. So there are two paths to keep investigating:

 - MapboxGL has some buried state in the click event that can indicate drawing mode
 - MapboxGL Draw does have some events that tell us exactly when drawing has started and drawing has ended, and that the drawing ended event occurs before the MapboxGL click event (it would have to set some state before checking again).

allthesignals:
See #311 - still remaining issues with route transitions.

chriswhong:
Taking a crack at this without using the default draw controls.  I also figured out a hack to get around the bug that draw intercepts touch events, which I'll try to incorportate

PR #313, opened 2017-10-12
Fade transitions for lots

allthesignals:
This is an experiment with fade transitions between lots.


andycochran:
There were conflicts with develop. I've fixed them. But the tests are still failing. 

There's one awkward behavior I notice: when you change between tax lot routes, the new content loads before the fade out/in. Maybe setting the fade out timing to zero would fix this?  

If we do accept this PR, maybe it should apply to all route transitions, not just lot-to-lot. 

But I'm not certain this is necessary. 🤔

andycochran:
Closing this for now. We can revisit if it's prioritized in the future or if it closes an issue brought up by customer/users.

PR #317, opened 2017-10-16
Redirect if root/no hash

### LINKED ISSUE #225: Default App State
Several folks have mentioned wanting the About page — which is actually currently a welcome message — to load as the default landing page. This can also be the location for the feedback links for #208. 

A method for doing this was mentioned in #72. But that issue resolved 'cause it was focused on a separate (tho related) UX. 

I propose we approach this the way  [this Shareabouts app](http://ideas.pbnyc.org/) does: 

When you navigate to `/` (with no hashed map parameters), we redirect you to `/about`. When you close the content area on any route, we route to `/#[zoom]/[lat]/[lng]`. 

What do y'all think of this, @allthesignals @chriswhong?

allthesignals:
Redirects to about page if no hash is currently in the route. Eventually, a hash is appended, even if there is none in the original route. Closes #225

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -8,8 +8,9 @@ const { service } = Ember.inject;
 export default Ember.Route.extend({
   mainMap: service(),
 
-  beforeModel() {
-    if (!window.location.hash) {
+  beforeModel(transition) {
+    console.log(transition);
+    if (transition.intent.url === '/') {
       this.transitionTo('about');
     }
   },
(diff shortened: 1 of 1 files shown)


andycochran:
This works really well. 

I did notice that if you refresh the root route it always goes to About. Even with a hash it redirects to About — e.g. `/#10.2/40.7071/-74` redirects to `/about#10.2/40.7071/-74`. 

I don't think this is a problem; I like this behavior. Only hard refreshes cause a redirect with or without a hash. I just wanted to note it in case it's not intended.

allthesignals:
@andycochran cool - that behavior is intentional but can be changed by checking for a hash in the window location (currently it checks to see the route "intent" or whether any child routes are present).

PR #322, opened 2017-10-17
Fruchala text changes

andycochran:
This PR… 
- [x] Updates the descriptions of zoning districts (per @fruchala) 
- [x] Adds a `zoningAbbr ` crosswalk to look get an abbreviated zoning district, 
- [x] Uses the abbr for the description crosswalk (instead of `primaryzone`, which needs to stay separate, as it's used for external links)
- [x] Adds an R10 checkbox to the palette 
- [x] Adds a darker yellow for R10 (the spectrum was previously only R1-R9)
- [x] Fixes the awkward spacing in the `zoning-district.hbs` template

There are still a few bugs…
- [ ] R10s are styled like R1s. R1 thru R9 fade from light to dark yellow. R10 is light like R1. (It's like this in `develop`)
- [ ] You can't filter the map by R10. When it's the only checked Residential District filter, it doesn't show on the map. When all others except it are checked, it still shows.

### DIFF
--- a/app/controllers/application.js
+++ b/app/controllers/application.js
@@ -48,6 +48,7 @@ export const mapQueryParams =
         R7: { defaultValue: true },
         R8: { defaultValue: true },
         R9: { defaultValue: true },
+        R10: { defaultValue: true },
         c11: { defaultValue: true },
         c12: { defaultValue: true },
         c13: { defaultValue: true },
--- a/app/layer-groups/zoning-districts.js
+++ b/app/layer-groups/zoning-districts.js
@@ -37,6 +37,7 @@ export default {
               ['R7', '#f3f63e'],
               ['R8', '#f3f62b'],
               ['R9', '#f2f618'],
+              ['R10', '#F0F614'],
             ],
           },
           'fill-opacity': {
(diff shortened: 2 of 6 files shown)


allthesignals:
The SQL query template uses this formatting:
`SELECT * FROM (SELECT *, LEFT(zonedist, 2) as primaryzone FROM support_zoning_zd) a WHERE primaryzone IN ('R10')`

The `LEFT(zonedist, 2)` function truncates the string up to the number of characters in the second argument... so it's looking for R1 and R1.

andycochran:
This concept of "primary zones" is a lot of trouble — both in the sql as you mention, and in the descriptions crosswalk — because of the inconsistent combinations of numbers, dashes, and letters.

andycochran:
I wonder if it'd be better to just query the data without truncating, and keep a separate table or crosswalk in the app to look up the "primary zone" when necessary…

allthesignals:
Agreed... I'm gonna look this over a little bit more.

allthesignals:
@andycochran take a look

andycochran:
@allthesignals Documenting our IRL discussion… 

Instead of `LEFT(zonedist, 2)`, the sql needs to get the first 2 or 3 characters: 
- Select the first 3 characters if the third is a number (you can get `R10`)
- Select the first 2 characters if the third is a dash or letter (so `R3-1` gives you `R3` and `R4B` gives you `R4`)

allthesignals:
@andycochran please merge when you think it's good to go. Otherwise, lemme know. Thanks.

andycochran:
I think this is ready. Waiting on test to pass then I'll merge. 🎉

PR #329, opened 2017-10-19
Link to intersecting

allthesignals:
Adds new component, link-to-intersecting, which performs a spatial query against a specific table and geometry.


allthesignals:
@andycochran yeah, it's a poorly-named component. It's really just fetching some data with a given model and tablename, and assumes a bunch of stuff about Carto. In the external link scenario you raise, I have it yielding returned data only if there's an intersection that occurs. So external links will work like this:

```

andycochran:
This looks great, @allthesignals! 

I made one change, moved the Tax/Zoning map links below these new links. I think these intersecting layers make more sense right below the HistDist and ComOverlays, whereas the map links are a different type of thing (maybe they get moved?). 

If you think this is ready, let's merge it. 🎉

allthesignals:
Thanks! All done

PR #335, opened 2017-10-20
Add active style to Mapbox GL Draw button

andycochran:
This PR indicates to the user that they are in drawing mode by changing the color of the draw button. 

However, because clicking the draw button a second time does not cancel drawing mode, there's an [open issue in Mapbox GL Draw](https://github.com/mapbox/mapbox-gl-draw/issues/698) upon which this PR depends.

### DIFF
--- a/app/styles/modules/_m-maps.scss
+++ b/app/styles/modules/_m-maps.scss
@@ -133,6 +133,10 @@
 }
 
 // Drawing
+.mapbox-gl-draw_ctrl-draw-btn.active {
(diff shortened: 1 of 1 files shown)


chriswhong:
Closing for now, as we no longer have mapbox-gl-draw in the app.  Can re-open if we choose to implement measurement

andycochran:
This also may not be relevant if we choose to create a custom UI (as we did for ZoLa).

PR #355, opened 2017-10-23
341 intersecting layers refactor

### LINKED ISSUE #341: Add Mandatory and Voluntary Housing Designated Layers in Intersecting Map Layers
It would be good for people to be able to see at a glance how their lot is affected by particular supporting zoning. It's more important than the others.

In this issue, we should also refactor the way intersecting layers are handled. Currently, they are not tasks and there's no way to round up the status of all promises. 

Maybe this should be done in the route view? It will slow the overall load time of the intersecting layers calculations.

allthesignals:
This PR does three things:

 - Refactors intersecting-layers component to yield a resolved hash of keys and values that are truthy when an intersection is found. 
 - Refactor intersection calculation into restartable concurrency task to eliminate race conditions
 - As mandatory inclusionary housing into layer intersection check

@andycochran, the Zoning Districts section still uses link-to-intersecting... do I need to swap out those too? For the "none found" logic? Lemme know ur design input. Thanks. 

Closes #341


andycochran:
Wait. How is `link-to-intersecting` still working if you've removed the component?

allthesignals:
@andycochran because I reverted the commit that removed it

andycochran:
Shouldn't they all just use `intersecting-layers`?

allthesignals:
@andycochran probably, yeah - I think that's best. `link-to-intersecting` includes some logic for creating a link-to component, but it's probably redundant now.

andycochran:
Cool. Let's nix it. Wanna make that change then I'll take another test drive?

allthesignals:
@andycochran will do!

PR #399, opened 2017-11-20
Cw various

### LINKED ISSUE #255: User sees grab cursor (not pointer) for features that aren't clickable.
Related to #251.

### LINKED ISSUE #325: Map center off in default app state
When the index page redirects to About, we should set padding on the right side of the map so it doesn't look like Newark’s Zoning & Land Use Map.

image

### LINKED ISSUE #367: Remove 'Shaded X' 500 Year floodplains & Revise 2015PFIRM Tooltip
The 'Shaded X' 500 Year floodplain doesn't really have any significance for zoning and most other regulatory purposes and should be removed from both 2007 and 2015 FIRM maps.

2015 PFIRM tooltip should read "“Released in 2015 as part of a citywide flood map update, the Preliminary FIRMs establish the 1% annual chance floodplain. For building code and zoning purposes, the more expansive of the either the 2015 or 2007 maps is used.”

### LINKED ISSUE #372: Supporting Zoning Layers - additional info
Previously, clicking on supporting layers would bring up the right hand panel with links to outside information. It is no longer coming up for a number of layers (MIH, Inclusionary, Transit Zone, FRESH, LDGMA, Coastal Zone, WAP, historic districts, flood layers).

### LINKED ISSUE #386: lot view redirect
Matt Croswell: "The Zoning District link on the lot view redirects to a citywide view of the zoning district instead of the info page. It is not helpful to see where all R8As are when I’m looking at a specific lot I searched. The citywide view can be achieved by doing a general search on a zoning district and has a link from zoning district page."

chriswhong:
This PR:
Closes #386 
Links to zones in the lot view go directly to DCP instead of the App's internal route for zones
Closes #372 
Moves special purpose districts and subdistricts up in the layer palette into the zoning and land use group
Closes #325 
Adds fitbounds to the default view so that the map is centered properly with the about tab open on large screens.
Closes #367
Removes shaded X zone from floodplain sources, layers, and legends
Closes #255 
Sets cursor to pointer only if highlighted layergroup is also clickable


andycochran:
External links need the external link icon.

Issue #309, opened 2017-10-11
User can click on, search for, route to Commercial Overlays. Details link to handbook info. 


andycochran:
Need to get some summary text so we can display it on the route.

andycochran:
Use the tooltip for the summary text. Unblocking…

andycochran:
@allthesignals I think we should close this as wontfix. You can already get to Commercial Overlay details from the lot view. We need to establish rule for why something is routable. I don't think "so you can link externally" is a good reason for something to be routable. Special Purpose districts are routable because you can search for them, which is a good reason. Objections to closing this?

fruchala:
It’s weird to me that people would be able to search for nearly every zoning district except for C1 and c2 overlays which are some of the most important in the ZR. Why this one as opposed to any of the others?

Sent from my iPhone

On Oct 20, 2017, at 10:02 AM, Andy Cochran <notifications@github.com<mailto:notifications@github.com>> wrote:


@allthesignals<https://github.com/allthesignals> I think we should close this as wontfix. You can already get to Commercial Overlay details from the lot view. We need to establish rule for why something is routable. I don't think "so you can link externally" is a good reason for something to be routable. Special Purpose districts are routable because you can search for them, which is a good reason. Objections to closing this?

andycochran:
Thanks, @fruchala. That's a good reason for making COs routable AND searchable.

allthesignals:
How should routing working for this? Is it feature-level or...?

allthesignals:
![Screen Shot 2017-11-20 at 1.48.50 PM.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5830e3bc40b5a00e00a5b4c0/125516c66e82c728ace21e0d46db978826878de9bce1b057e40fa0c57e0b76226456f36b7191a379f116234f5a5f5ecc3c4a4356bcbbd671f3e57862804e0ee4836d5b7d903a17a9ea8149e901ba21d94684b7eddb4dd2f42f67b6d5db1063474f4dcbed5727a29941c95df74ae34d56a3173e6ae47e2431.png)

Any thoughts on the language for this view? I'll just link to commercial overlay?

PR #403, opened 2017-11-20
374 co rh col

### LINKED ISSUE #374: Commercial Overlays - Right hand column
Commercial overlays are getting lost in the right hand panel in the lot info panel. This is conflicting with how other applicable zoning districts are treated.

allthesignals:
Depends on #309, re-adds link to commercial-overlay. Closes #374


andycochran:
@fruchala @allthesignals Now that I see this in the app, I think we should change the text. If we're not using the fill patterns in ZoLa, we shouldn't list them there. Sorry I didn't think this through before.

allthesignals:
@andycochran no prob - what do you propose?

andycochran:
I'll edit this copy then merge…

PR #408, opened 2017-11-21
Data Page and Sourcing

### LINKED ISSUE #236: Cite Data Sources
Cite and Link to data sources used by this application. 

Recommend this live in a markdown file in version control, and can be linked to from /About

chriswhong:
This is a placeholder PR for the new data route and layergroup MetaData.  Layergroups now have a `meta` property with `description`, `url`, and `created_at` as children.

Still TODO:
- [x] @andycochran can you take a look at the layout?  A simple list is all we need, but the colors and sizing need some work.
- [x] Add a new "Data" button to the header 

Closes #236


andycochran:
I've added formatting/style. Looks like there are a few with incomplete data. I also made the markup conditional so the formatting won't break if meta info is missing.

andycochran:
Did I break the tests?

andycochran:
Oh. Looks like my push to this branch failed. But I saw the broken tests and thought it had pushed. I'll merge those style changes separately…

PR #438, opened 2017-12-05
Add /bbl route

pichot:
Simple commit that adds a `/bbl` route. Uses already implemented `bbl-demux` to parse bbl and redirect to `/lot` route.

Unit and acceptance tests included.

### DIFF
--- a/app/routes/bbl.js
+++ b/app/routes/bbl.js
@@ -1,0 +1,9 @@
+import Ember from 'ember';
+import bblDemux from '../utils/bbl-demux';
+
+export default Ember.Route.extend({
+  model(params) {
+    const bbl = bblDemux(params.bbl);
+    this.transitionTo('lot', bbl.boro, bbl.block, bbl.lot);
+  },
+});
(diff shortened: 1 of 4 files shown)


app/routes/bbl.js
```diff
@@ -0,0 +1,9 @@
+import Ember from 'ember';
+import bblDemux from '../utils/bbl-demux';
+
+export default Ember.Route.extend({
+  model(params) {
```
pichot: @allthesignals is this the right hook to use? I tried `beforeModel` but it appeared that the params weren't fully loaded.

pichot:
Rebased and ready to merge please :)

app/routes/bbl.js
allthesignals: @pichot yes - beforeModel doesn't have the params object directly, you'd have to access it differently.

app/routes/bbl.js
```diff
@@ -0,0 +1,9 @@
...
+    const bbl = bblDemux(params.bbl);
```
allthesignals: You might also have said:

```javascript
export default Ember.Route.extend({
  model({ bbl }) {
    const { boro, block, lot } = bblDemux(bbl);
    this.transitionTo('lot', boro, block, lot);
  }
});
```

pichot: oooh, better. Should have waited for your code review

PR #463, opened 2017-12-27
375 e designations

### LINKED ISSUE #375: E designations
Need to bring back E designation layer

chriswhong:
This PR adds the environmental designations layer group.

Changes Proposed:
- E-designations appear as a blue circle with an "E" in the center
- Clicking an e-designation will open a LUCATS link using the `ceqr_num` property.  Not all of these are valid, and those that are not will just show a LUCATS search page.

Closes #375


app/controllers/application.js
```diff
@@ -120,7 +120,22 @@ export default Ember.Controller.extend(mapQueryParams.Mixin, {
...
+            sdlbl,
+            splbl,
+            overlay,
+            cartodb_id,
+            ceqr_num,
+          } = feature.properties;
+
+          //  handle e-designation click
+          if (ceqr_num) { // eslint-disable-line
+            window.open(`http://a030-lucats.nyc.gov/lucats/DirectAccess.aspx?CEQRNO=${ceqr_num}`) // eslint-disable-line
```
allthesignals: @chriswhong you should add a little defensive programming with the `window` object in case it does not exist for node-based environments for whatever reason `window !== 'undefined' && window.open`. 

In terms of UX, does it make sense to send users off-site after clicking on the map? Thoughts?

app/layer-groups/subway.js
```diff
@@ -5,8 +5,8 @@ export default {
   meta: {
     description: 'NYC Subway Lines and Stops - Originally Sourced from NYC DoITT GIS, combined with SI Railway data from Baruch College NYC Mass Transit Spatial Layers | Subway entrances from NYC Open Data',
     url: [
-      'https://planninglabs.carto.com/api/v2/sql?q=SELECT * FROM support_trans_mta_subway_stops',
-      'https://planninglabs.carto.com/api/v2/sql?q=SELECT * FROM support_trans_mta_subway_routes',
+      'https://planninglabs.carto.com/api/v2/sql?q=SELECT * FROM support_trans_mta_subway_stops&format=SHP',
+      'https://planninglabs.carto.com/api/v2/sql?q=SELECT * FROM support_trans_mta_subway_routes&format=SHP',
```
allthesignals: @chriswhong curious, what's this for? Separate issue?

app/controllers/application.js
chriswhong: It's either that or we add a route for e-designations.  I don't think there's anything wrong with linking out directly from a map click as long as it opens a new window/tab

app/layer-groups/subway.js
chriswhong: Yes, it's the only other meta with a carto API call, and it should be a SHP download, not a CSV.

app/controllers/application.js
allthesignals: Okay - our app doesn't do this anywhere else and external links are always clearly marked. I think it's awkward, but it's your call.

chriswhong: I killed it

allthesignals: Sorrrrryyyy

PR #465, opened 2017-12-27
set building footprints on by default

### LINKED ISSUE #448: Have building footprints on by default
I always want to see the building footprints – could it be on by default? Or have an option (perhaps with a cookie?) that remembers that I want it on by default?

chriswhong:
This PR turns the building footprints layergroup on by default

Closes #448


allthesignals:
Great - just need to rebase/merge with develop

PR #472, opened 2017-12-27
Working browser compatibility warning

### LINKED ISSUE #452: Graceful degradation / failure
#### User Story
As a user, I want to know what browser to use so that I can use this!

#### Description
We need to implement OS / Browser detection so that we can redirect the user to a helpful message when a known, unsupported Browser / OS combo is being used. Currently, this does not exist, and it should be a standard for our apps.

allthesignals:
This adds an aggressive overlay about unsupported browsers. We should test this thoroughly in Browserstack. 

Closes #452


allthesignals:
We should merge once @andycochran has some input on how to approach this...

In the meantime, we should create a compatibility matrix to understand where issues are catastrophic (no map!!!) or merely annoying (CSS glitches).

PR #476, opened 2017-12-28
Use acris query params

### LINKED ISSUE #456: ACRIS links should link to ACRIS BBL profile
Received via email:  The ACRIS link does not take you to the property’s page like it does in DoITT’s NYCity Map.  Instead you have to search for the property again once you arrive at the ACRIS page. It would be helpful if the link in the ZOLA page would automatically bring up a BBL’s ACRIS profile.

allthesignals:
This PR changes the lot template to use the acris query-param-enabled URL to pull up the BBL-specific link.

Closes #456


chriswhong:
Looks great, not sure we we implemented it the forms-y way to begin with, but glad this was an easy fix.

PR #536, opened 2018-03-21
Feature/378 select multiple lots

### LINKED ISSUE #378: Allow Selection of Multiple Tax Lots
Carolyn Grossman writes: "I couldn’t figure out how to pull up multiple tax lots at once, which was a cool feature of old version. Seemed like you would have to pull and save each lot, then export out to get a list." 

Frank says - This seems to be a feature to consider adding later. I had no idea people used ZOLA like this.

chriswhong:
This PR implements a "highlighted" symbology for saved tax lots.   It will also highlight the lots when hovering over the corresponding list item in the "saved" pane.

We'd love to get some feedback from @fruchala and @croswell81 on this before merging.
Closes #378


croswell81:
The symbology for the saved lots should be toggleable so user can turn the layer on/off.

PR #537, opened 2018-03-21
Feature/502 search by block

### LINKED ISSUE #502: Add ability to search by block
#### Description
The previous ZoLa had an ability to search by tax block. This was very helpful, as users could search by blocks when lot numbers are unavailable. It would be helpful if this were added as a feature.

allthesignals:
This PR adds functionality to search for block. If a block is found, the map will fly to the centroid. Otherwise, an error will result.

Closes #502

@andycochran @chriswhong lemme know thoughts on functionality - should there be a route or even just an outline of the block?


chriswhong:
Can I suggest that we validate the input with each keypress and disable the button if a valid block and lot is not selected?

It requires more API calls but I think it might be a better UI.

allthesignals:
@chriswhong yes, I replied in the issue - I was hoping for that level of specificity up front. I think that added complexity would be another issue though. Should this PR not be merged until we split this off?

Inck:
I'm seeing incorrect centroids load across a few test addresses when running this branch locally. Example: 

Lot:
http://zola.planninglabs.nyc/lot/3/2654/37

Block:

screen shot 2018-03-22 at 12 20 39 pm

chriswhong:
Let's revisit this functionality while addressing #574

chriswhong:
No longer relevant, BBL lookup was ported in from streets app in #632

PR #542, opened 2018-03-22
Set a default land use color so "Other" isn't so prominent

andycochran:
This PR adds a `default` parameter to the `fill-color` on the `pluto-fill` layer. This gives the "Other" land use a specific color, instead of defaulting to the stock Mapbox style (black). 

### Before: 
image

### After: 
image


allthesignals:
👏

croswell81:
Yeah, the white is much better than the black

Issue #441, opened 2017-12-06
Add Historic Aerial Raster Layers


fruchala:
Request to add in older aerial photos (1924, 1996, etc) into ZoLa. were not available in previous version

chriswhong:
Various historic imagery layers are published by DoITT GIS here: https://maps.nyc.gov/tiles/

However, the current aerials are proxied through a small node service (originally setup for the capital planning platform and used in ZoLa), I believe this was due to CORS issues with the published services.

chriswhong:
Prioritize: 1924, 1951, 1996

Implied in this issue is ensuring that the user cannot turn on more than one raster layer at a time, so turning on an aerials layer should turn off all others.

If possible without too much more effort, add all available.    
Idea: to avoid the layer palette getting very long, add a dropdown to the Aerials layer to choose the year.

allthesignals:
Layers and endpoints:
 - [ ] 1924 
 - [ ] 1951 
 - [ ] 1996

allthesignals:
@chriswhong 

> However, the current aerials are proxied through a small node service (originally setup for the capital planning platform and used in ZoLa), I believe this was due to CORS issues with the published services.

Is this a service we maintain? Is it on Github somewhere w/ a readme? If the layers don't support CORS, we'll have to use a proxy.

chriswhong:
@allthesignals this is not the case anymore, it was about a year ago, but the DoITT tiles are now usable directly as they are published.

allthesignals:
How should this be styled? @chriswhong @andycochran 
<img width="321" alt="screen shot 2018-03-12 at 2 37 00 pm" src="https://user-images.githubusercontent.com/5004319/37302771-e5338c02-2602-11e8-8c14-06a93990b4db.png">
<img width="163" alt="screen shot 2018-03-12 at 2 36 54 pm" src="https://user-images.githubusercontent.com/5004319/37302772-e544e736-2602-11e8-9db7-cf408c7a24c9.png">
<img width="169" alt="screen shot 2018-03-12 at 2 36 45 pm" src="https://user-images.githubusercontent.com/5004319/37302773-e550b6e2-2602-11e8-8373-be6f19df76e1.png">
<img width="331" alt="screen shot 2018-03-12 at 2 36 39 pm" src="https://user-images.githubusercontent.com/5004319/37302774-e55dd0f2-2602-11e8-8b13-1dd0888c7374.png">

chriswhong:
[...]

PR #552, opened 2018-03-29
Cw aerials improvement

### LINKED ISSUE #545: Historic Aerials are out of order

chriswhong:
This PR
- adds all aerial raster layers published by NYC DoITT
- changes the icon to use when an aerial layer is selected (now using `fa-dot-circle-o`)
- adds query param support for individual aerial layers
- adds a `radio-selector` component (child of `layer-menu-item` to show a radio button UI for choosing from the aerials layers in the layer-group

Closes #545, improves #441


app/components/radio-selector.js
```diff
@@ -0,0 +1,58 @@
...
+      const layers = this.get('layers');
+      const qps = this.get('qps');
+
+      const matchedLayer = layers.find(layer => qps.get(layer.layer.id) === true);
+
+      if (matchedLayer) {
+        this.send('switchLayer', matchedLayer.layer.id);
+      }
+    });
+  },
```
allthesignals: We talked a bit about this larger design issue - overriding init and conditionally determining an initial state is something we do elsewhere, but I wonder if it's a code smell? Basically, we have query params, which have defaults that are determined by configs, but query params are allowed to be set by the URL. Basically, I wonder if this component can just be responsible for managing the layer visibility state, and this overriding behavior happens somewhere else higher up in the app.

Issue #479, opened 2017-12-29
Validate BBL Lookup


allthesignals:
#### What should happen:
When BBL Lookup form is incomplete, it should not be possible to hit submit.

#### What happened instead:
Hitting submit with incomplete BBL Lookup form returns a 400 from Carto.

#### How to reproduce this bug:

1. Select Boro
2. Select Block
3. Leave Lot Blank
4. Hit submit

#### Browser(s) and Device(s) observed on:

See TrackJS for this: https://my.trackjs.com/shared/ZTY2ZjQ1MTE5ZmJmNDZlOWE2ZTY0NGQyODc1ZjkxNDI.

chriswhong:
Related to #502

chriswhong:
A nice UI for this was added to streets.planning.nyc.gov, we can port it back to ZoLa.

PR #624, opened 2018-07-31
Improve BBL Form UI

### LINKED ISSUE #574: BBL Look Up Search should initiate by pressing 'Enter' button
The BBL Look Up search does not recognize the enter key to initiate block lot search. You always have to click the 'Go to BBL button' 

(from user: tolve@earthlink.net)

andycochran:
This PR closes #574 by…
- Changing the `div` to a `form` so that pressing enter in an input triggers the submit action
- Adds actions to the inputs so that changing them resets the error message to `''`

### DIFF
--- a/app/components/bbl-lookup.js
+++ b/app/components/bbl-lookup.js
@@ -42,8 +42,13 @@ export default Component.extend({
       });
     },
 
+    resetErrorMessage() {
+      this.set('errorMessage', '');
+    },
+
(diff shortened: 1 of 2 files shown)


andycochran:
I'm closing this, because I don't think it's the right way to go. #502 and #479 require porting over City Map's BBL lookup, and I think this should be resolved with/after that.

PR #625, opened 2018-07-31
refactor computed that applies saved class to bookmark button

### LINKED ISSUE #618: Bookmark Button does not turn green when clicked
**Describe the bug**
Bookmark Button does not turn green when clicked.  

**To Reproduce**
Navigate to a lot route.  Click the bookmark icon.  The lot will be "saved", and is highlighted on the map and appears in the bookmarks pane, but the bookmark icon does not animate, turn green, and show "Saved" like it used to.

**Expected behavior**

chriswhong:
This PR updates the computed property that applies a `saved` class to the bookmark button.

Closes #618


allthesignals:
I'm having trouble reproducing this bug in the current develop branch

PR #671, opened 2018-10-15
Install labs-ember-search

allthesignals:
Refactor to use search API 
Refactor tests
Make test runners use GPU in headless Chrome

This fixes an issue in production in which keying down the list of search results highlights results of two types.


allthesignals:
@andycochran could you take a look at this styling problem? I'm going to look at the highlighting problem. Thanks @hannahkates

allthesignals:
@hannahkates I'm having trouble reproducing

allthesignals:
@hannahkates I can't seem to reproduce this what steps did you take?

hannahkates:
@allthesignals 
0. Go to https://zola-staging.planninglabs.nyc/
1. Type "120 bro" in the search bar. You'll see that "120 Broadway, Manhattan" is highlighted in yellow.
2. Mouse down further on the list, and you'll see that each result you hover over will be highlighted, but "120 Broadway, Manhattan" continues to be highlighted even when you're not hovering over it anymore.

As a user, I don't actually find this to be alarming/confusing, but I think the PR was aimed at fixing this multiple highlighting issue.

andycochran:
So, while it is possible to highlight two search results—if you arrow down to highlight a specific result, you can mouse over another—I don't think that's a bug, as you're using two different methods to highlight something. @hannahkates, is that how you're highlighting multiple records? I'm going to go ahead and merge this PR. Please open a separate issue for the highlight bug if you can define steps to reproduce. We _could_ style hovering differently than arrowing, but that should also be a separate issue.

allthesignals:
> So, while it is possible to highlight two search results—if you arrow down to highlight a specific result, you can mouse over another—I don't think that's a bug, as you're using two different methods to highlight something. @hannahkates, is that how you're highlighting multiple records?

Yes, to your point, Github has the same issue haha:
screen shot 2018-10-23 at 10 36 34 am

Also, @andycochran, @hannahkates pointed out that the magnifying glass wasn't centered correctly... are you not seeing that anymore?

hannahkates:
I was using the mouse, not the arrow key, to highlight the records below the first highlighted result. The magnifying glass looks to be fixed now.

allthesignals:
@hannahkates thanks! Yeah, the original issue was that it was highlighting rows across multiple result _types_, which was wrong. Thanks everyone.

PR #690, opened 2018-10-26
680 init map to lot

### LINKED ISSUE #680: Direct to route navigation triggers dozens of expensive tiles
Navigation to a direct route (a lot view, for example)) triggers an animated map transition which fires a torrent of expensive network requests for new tiles (that are even used). We should find a way to set the initial map view from the route.

allthesignals:
This PR closes #680 by eliminating the `fitBounds` animation when the routing intent is something other than application, about, or index, reducing the average lot profile view size to around 2mb (down from 6mb). 

This PR also refactors the way fitBounds is handled. Instead of an awkward template-based approach, we explicitly call `fitBounds` from the map instance. We handle asynchrony by using tasks that wait for the mapInstance to be available. 

Please pay special attention to these routes:
- [ ] lot
- [ ] zma
- [ ] bbl
- [ ] zoning-district
- [ ] special-purpose-district
- [ ] special-purpose-subdistricts
- [ ] commercial-overlay

Manual test them by:

1. Turning on their respective layer if needed.
2. Navigating to that layer by clicking on it from the map.
3. Checking that there's no missing content (telltale signs include lone `|`s, templates with only generic text).
4. Reload the route and make sure that it fits the map to the bounds of the resources. Lot route should _not_ animate, others do. 
5. Zoom around on the map, then, if present, click the button "zoom this all [resourceName]". This button is sometimes visible at the bottom of the profile view.

### DIFF
--- a/app/components/main-map.js
+++ b/app/components/main-map.js
@@ -50,7 +50,7 @@ class MainMap extends Component {
 
   lat = 40.7125;
 
-  @computed('mainMap.isSelectedBoundsOptions')
+  @computed()
   get lng() {
     const boundsOptions = this.get('mainMap.isSelectedBoundsOptions');
     return boundsOptions.offset[0] === 0 ? -73.9022 : -73.733;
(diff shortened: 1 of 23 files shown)


allthesignals:
@hannahkates should be good now

Issue #698, opened 2018-11-27
Clean up unused code from before labs-layer-api was implemented


hannahkates:
There are relic layer source and group files for the zoning layers used in ZoLa. We should delete these.

allthesignals:
Yes - I've been doing some of this by deleting services. The immediate ones are:

 - [ ] layers folder
 - [ ] sources folder

PR #720, opened 2018-12-12
Resolve qps

### LINKED ISSUE #682: New staging: Icon in search bar isn't centered
This is happening on Chrome on my macbook, not connected to additional monitors.

Screenshot:
image

### LINKED ISSUE #687: New staging: Special purpose subdistrict profiles don't load
Steps to reproduce:
1. Turn on special purpose subdistricts
2. Turn off other layers so it's easier to click on special purpose subdistricts
3. Click on a special purpose subdistrict
4. Profile pane appears, URL updates, but profile content won't load

### LINKED ISSUE #688: New staging: Special purpose districts aren't clickable

### LINKED ISSUE #711: Rewrite grouped checkboxes
Related to #705

Currently: grouped checkboxes use a strange system of storying lists of key names (C1, C2, etc), and using those names to lookup and change state to filtering. This is a bit awkward and does not currently work with mapbox composer layer group models. 

screen shot 2018-12-01 at 3 48 19 pm

We need to rewrite this functionality in a cleaner way - it does not really work reliably right now! What this involves is creating a translator interface between *MapboxGL's filter specification* and *groups of checkboxes*. 

If you look at `layer-multi-select-control`, you will sen action that translates a list of "values" (things currently selected) into a mapbox-gl filter:

```javascript
  @action
  selectionChanged() {
    const values = this.allChecked;
    const { layerGroup, layerID, column } = this;

    const previousValues = this.values;

    if (JSON.stringify(values) !== JSON.stringify(previousValues)) {
      // should have access to the model and call the filter method
      const expression = ['any', ...values.map(value => ['==', column, value])];
      layerGroup.setFilterForLayer(layerID, expression);
    }

    this.set('values', values);
  }
```

This works for now, but has been unreliable. Sometimes the checkboxes don't trigger updates to the map. 

Using `selectionChanged` as a starting point, we should rewrite the components for these grouped checkboxes (see the screenshot).

### LINKED ISSUE #726: pending zoning map amendments appear to link to zmas
Clicking PZMAs routes to ZMAs, sometimes landing on wrong geometries

allthesignals:
This PR…

…

Closes #682 
Closes #687 
Closes #688 
Closes #711 
Closes #726 

Addresses #698


andycochran:
If you have qps on the index route (other routes seem okay) and refresh, you just see a white screen and two errors:
- `Error while processing route: about params.any is not a function TypeError: params.any is not a function`
- `Uncaught TypeError: params.any is not a function`

hannahkates:
@allthesignals Can you add a high level recap of what's included in this PR in the description? @trbmcginnis is going to work on issue #738 and I want to make sure there isn't any overlap.

allthesignals:
@hannahkates the 3 linked issues can cover that

julialucyhogan:
> `params: "[\"building-footprints\",\"subway\",\"tax-lots\",\"zoning-map-amendments\"]"` <- params are being _set_ but not parsed for some reason... they may be coming in to the application context before they work through the ember-parachute stuff (which handles serialization)

I'd be interested to know how you resolved this issue just for my own education :)

allthesignals:
> > `params: "[\"building-footprints\",\"subway\",\"tax-lots\",\"zoning-map-amendments\"]"` <- params are being _set_ but not parsed for some reason... they may be coming in to the application context before they work through the ember-parachute stuff (which handles serialization)
> 
> I'd be interested to know how you resolved this issue just for my own education :)

@julialucyhogan, well, unfortunately I could not understand why ember-parachute - an addon we use to improve Ember's core query-parameter handling - was not correctly parsing the stringified JSON coming from the URL during a route transition. So, before passing that information to a service that coordinates state across the layer models, UI, and URL, I parse it using `JSON.parse`, if it comes down as a string.

If it comes down as a string and it's not proper JSON, the app will crash, so it's not my favorite solution. This scenario should not happen because the feature wasn't designed to handle users who type in additional things into the URL (but get it wrong). Perhaps this could be added with a simple try, catch, since it's a little flaky.

allthesignals:
> The fitBounds function is behaving strangely in this PR. Instead of the map displaying the selected feature in the center of the map, the selected feature overlaps with the legend pane on the left. See gif example below:
> 
> zola fitbounds issue

This is happening on staging, though, as far as I can tell. I need to get this merged so I'm going to dismiss for now and address in a later PR.

PR #725, opened 2018-12-13
[WIP] Update layer click

### LINKED ISSUE #738: Make layer interactivity consistent
All interactivity should match that's laid out in this Airtable: https://airtable.com/tbl7qrEzVgOlHsqBR/viw85aFEf3ld7KFa2

trbmcginnis:
- Update `update-selection.js` mixin by removing fitbounds for clicking on zoning district layers (decided that this was too distracting for users because map constantly moves around when user clicks on new zoning district)
- Add placeholders on `zoning-district.hbs` so that users can see that model is loading

Closes #738.


app/mixins/update-selection.js
```diff
@@ -9,24 +9,11 @@ export default Mixin.create({
...
-    yield waitForProperty(taskInstance, 'state', 'finished');
-
-    this.get('mainMap.setBounds').perform();
-  }).restartable(),
-
-  actions: {
-    didTransition() {
-      this.get('mainMap.setBounds').perform();
-    },
-  },
```
allthesignals: So this is a mixin, meaning anywhere else the mixin is used will also lose the ability to zoom to selection. Where else is this used? Do we need to split this out into another mixin - one for fitting bounds and one for setting the seelction?

trbmcginnis: So from my understanding we have two ways to get to the zoning districts route:

1) **Clicking on the map layer**
expected behavior is that user clicks on layer, all layers with same zoning label change color, but map does NOT fitbounds
2) **Searching for a zoning district in the search bar**
expected behavior is that user selects search result from dropdown and ALL of zoning districts with same label change color and are selected, map "fitbounds" to ALL of districts that share that label

THEN, from the zoning districts route, there is a **button with "Fit map to all _____ "**, which fits the bounds to ALL of the zoning districts with that label (same behavior as search). There is an action on the route that is used for this button called `fitBounds`, which is separate from the mixin. 

So I think you're right that this mixin will affect SEARCH, because it will assume that going to the route WILL NOT fitbounds. So maybe it is a good idea to split it up? Or maybe if there's a way to override it for `handleLayerClick`?

allthesignals: > So maybe it is a good idea to split it up?

Yes... doesn't this mixin affect a bunch of other routes? It's not just about zoning districts - mixins can be used for _any_ route. If you search the source code for where the mixin is used, you can tell where else these changes will affect the app. 

So what does this mixin do? It appears to do too much.

I guess I don't quite understand why removing an entire method is the way forward on this because it's going to affect every single route that uses the mixin. Have you checked to see how it affects the other layer groups?

trbmcginnis: > Yes... doesn't this mixin affect a bunch of other routes? It's not just about zoning districts - mixins can be used for _any_ route. If you search the source code for where the mixin is used, you can tell where else these changes will affect the app.

The routes that are affected by this mixin are:

- commercial-overlay
- lot
- special-purpose-district
- special-purpose-subdistrict
- zma
- zoning-district

Originally (the version that is in Master right now), we did not have any sort of fitbounds in this mixin. I don't think we need fitbounds behavior on any of these layers when clicked, so changing the mixin was appropriate for each of these routes. 

I think the big issue is that we then have two expected behaviors when a user lands on a route whether they click the layer or search for the layer (which in the case of searching we want it to do the same action as that Fit Map to All button). So in the case of search, the mixin provides the incorrect behavior. In case of clicking, it gives us what we want.

PR #779, opened 2019-02-06
Fix misleading mouse cursor issue

allthesignals:
Closes #658 (NYCPlanning/ember-mapbox-composer#31) by updating composer version which includes handling for “clickable” layer groups


allthesignals:
@julialucyhogan what should be the protocol for squashing merge conflicts? It seems weird to squash a commit message... I also get errors handling that. Thanks.

PR #781, opened 2019-02-06
700: 🖨 Print View! 

### LINKED ISSUE #700: Improve print functionality
- zoom to mapped center
- allow user to include the profile

Use case: DOB uses aerial imagery to confirm pre-existing land use exceptions

For reference, there's some old dialog about print use cases in #347

andycochran:
This PR adds support for a query parameter which adds classes in `application.hbs` to style the layout for printing using paper.css (+ many other custom print styles). The print view has loads of config options for paper size and orientation. WOOT! Better printing in ZoLa! Closes #700


julialucyhogan:
so awesome ! is it on the todo to remove the interactive-looking stuff (check boxes, drop down arrows) from the layer palette in print view ?

chriswhong:
Closes #700

trbmcginnis:
The cursor still shows up as pointer when panning the map when it should be grab hand.

allthesignals:
Need to add tests - there's only 1 test failing, which is for the new component. Gonna work on this today!

allthesignals:
@andycochran also plz prefix branch with issue # <3 thx

PR #813, opened 2019-03-11
807: Adds bbox route with validation

### LINKED ISSUE #807: Add `bbox` route that will fitbounds to bbox passed in via URL
Per discussion with HPD for compliance with URA legislation, we should add a new route to ZoLa that allows for programmatically building a ZoLa URL with a bounding box.  Once the app has loaded, the map will `fitBounds()` to the bounding box.

Proposed route: `/bbox/:west/:south/:east/:north`

What should it do if invalid values are passed in?  (Probably nothing, perhaps a console log, perhaps redirect to `/` route)

julialucyhogan:
closes #807

### DIFF
--- a/app/models/bbox.js
+++ b/app/models/bbox.js
@@ -1,0 +1,8 @@
+import DS from 'ember-data';
+import { attr } from '@ember-decorators/data';
+
+const { Model } = DS;
+
+export default class BboxModel extends Model {
+  @attr('array') bounds;
+}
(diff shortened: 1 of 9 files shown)


app/models/bbox.js
```diff
@@ -0,0 +1,8 @@
+import DS from 'ember-data';
+import { attr } from '@ember-decorators/data';
+
+const { Model } = DS;
+
+export default class BboxModel extends Model {
+  @attr('array') bounds;
+}
```
allthesignals: Nix it ! 😈

PR #814, opened 2019-03-12
Change `/about` redirect; keep zoom hash through clicks

chriswhong:
This PR:

- Hotfix: Only redirect to `/About` if the original url requested was `/`.  

- Hotfix: Add hack to ensure mapboxGL viewport hash remains after internal route transitions. (ensures that no matter where the user is in the app, the URL can be shared and will present the same map view.

### DIFF
--- a/app/router.js
+++ b/app/router.js
@@ -1,10 +1,23 @@
 import EmberRouter from '@ember/routing/router';
+import { inject as service } from '@ember/service';
 import config from './config/environment';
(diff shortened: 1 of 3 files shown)


julialucyhogan:
Some of the percy diffs appear to be real diffs, and I'm not sure exactly if they are intended. Please check them out. I've already approved all the noise-based no-op diffs.

Issue #841, opened 2019-06-03
Inject dependency of mapbox instance stub


hannahkates:

allthesignals:
https://github.com/NYCPlanning/labs-zola/pull/844

Issue #839, opened 2019-06-03
Wrap dependency components with your own layer


hannahkates:
Need to stub the map --> look for example in CEQR

allthesignals:
https://github.com/NYCPlanning/labs-zola/pull/844

Issue #838, opened 2019-06-03
Testing


hannahkates:
- Wrap dependency components with your own layer
- Stub layer
- Inject dependency of mapbox instance stub
- Check layer interactivity against airtable

Issue #815, opened 2019-03-15
Test that layer interactivity doesn't regress


hannahkates:
The layer interactivity should match the contents on this airtable: https://airtable.com/tbl7qrEzVgOlHsqBR/viw85aFEf3ld7KFa2?blocks=hide Let's set up tests to ensure we don't have regression during future refactors.

Build this table into test suite.

allthesignals:
https://github.com/NYCPlanning/labs-zola/pull/844

PR #844, opened 2019-06-06
838 Testing enhancements; mocks; stubs; offline testing; backfill

allthesignals:
This PR revamps the acceptance testing suites, and fills out tests for a major component of the application, the main-map component. 

A number of test support data factories are added to make testing easier. 

This PR addresses most if not all of the issues in the #838 epic... :

#59 Mock map search results for acceptance tests
#756 Mock Carto responses
#757 Mock Layers API
#815 Test that layer interactivity doesn't regress
#839 Wrap dependency components with your own layer
#841 Inject dependency of mapbox instance stub

I believe these are should be closed by this PR.


app/models/layer.js
```diff
@@ -0,0 +1,137 @@
...
+  @class LayerModel
+*/
+export default Model.extend({
+  init(...args) {
+    this._super(...args);
+
+    // enforce presence of blank object for mapboxGL validation
+    if (!this.get('style.layout')) this.set('style.layout', {});
+
+    // determine which is the first occurring layer
```
julialucyhogan: `// Set current layer as first occuring layer for owning layer group if one does not already exist`

app/models/layer.js
```diff
@@ -0,0 +1,137 @@
...
+    if (this.get('layerGroup') && !this.get('layerGroup._firstOccurringLayer')) {
+      this.set('layerGroup._firstOccurringLayer', this.get('id'));
+      this.set('position', 1);
+    }
+
+    this.delegateVisibility();
+    this.addObserver('layerGroup.visible', this, 'delegateVisibility');
+  },
+
+  delegateVisibility() {
```
julialucyhogan: I find if/elses hard to read, esp nested ones. I'm a fan of using returns to make the flow look nicer:
```
// Set visibility = false for singleton-type layers that are not at the top
if(this.get('layerVisibilityType') === 'singleton') {
  if(this.get('position') !== 1) {
    next(() => (!this.get('isDestroyed') ? this.set('visibility', false) : null));
    return;
  }
}
// set visibility = visible (from layerGroup) 
next(() => (!this.get('isDestroyed') ? this.set('visibility', visible) : null));
```

app/templates/components/labs-map.hbs
```diff
@@ -0,0 +1,47 @@
+{{!--
+  This template overrides the template for an addon, ember-mapbox-composer.
+  e-mapbox-composer uses private APIs in e-mapbox-gl. Those changed, so we need to
+  manually configure the interface here.
+--}}
+
+{{#if _loader.isLoaded}}
```
julialucyhogan: any reason not to collapse these into a single (and)-ed condition?

app/models/layer.js
```diff
@@ -0,0 +1,137 @@
...
+  */
+  tooltipTemplate: attr('string', { defaultValue: '' }),
+
+  paint: alias('style.paint'),
+  layout: alias('style.layout'),
+  layerVisibilityType: alias('layerGroup.layerVisibilityType'),
+
+
+  /**
+    Computed alias that returns a newly built mapbox layer object. Necessary to maintain state bindings.
```
julialucyhogan: what does "newly built" mean here?

app/models/layer.js
```diff
@@ -0,0 +1,137 @@
...
+    },
+  }),
+
+  /**
+    Getter and setter for visibility. Mutates a Mapbox property that actually determines visibility. Depends on parent visibility.
+
+    @property visibility
+    @type Boolean
+  */
+  visibility: computed('layout.visibility', {
```
julialucyhogan: perhaps should be named something more boolean-y? `isVisible `?

app/services/main-map.js
```diff
@@ -103,6 +101,7 @@ export default class MainMapService extends Service {
   @task(function* (explicitBounds) {
     const bounds = explicitBounds || this.get('bounds');
     while (!this.get('mapInstance')) {
+      console.log('running loop, map instance is falsey');
```
julialucyhogan: delete

config/environment.js
```diff
@@ -95,6 +95,9 @@ module.exports = function(environment) {
   };
 
   if (environment === 'development') {
+    ENV['ember-cli-mirage'] = {
+      enabled: false,
```
julialucyhogan: do you want to commit this change ?

app/templates/components/labs-map.hbs
```diff
@@ -0,0 +1,47 @@
+{{!--
+  This template overrides the template for an addon, ember-mapbox-composer.
+  e-mapbox-composer uses private APIs in e-mapbox-gl. Those changed, so we need to
```
julialucyhogan: does the new interface defined here also use private APIs?

app/models/layer.js
allthesignals: Sorry, this is code ported from an addon. I needed to fix something in the addon. This was pure laziness, I'll go ahead and make the fix in the addon, publish a new version, and update.

app/templates/components/labs-map.hbs
allthesignals: @julialucyhogan labs-map irresponsibly extends the mapbox-gl, but you cannot extend templates. This is ported from that addon (one our team does not own) so I brought it into our source code.

app/models/layer.js
allthesignals: My bad, I didn't explain properly - I would ignore this layer addition. I am going to make the fix in the addon from which it came,

config/environment.js
allthesignals: @julialucyhogan yes, because people tend to use `ember s` (develpment environment` oto check up on map styles, visual stuff.

app/templates/components/labs-map.hbs
allthesignals: hmm basically the old version of this component used some private apis in the addon ember-mapbox-gl. I updated ember-mapbox-gl, but it's private apis changed, as they do. so, we still use those private apis, but that should be addressed later.

allthesignals:
> This refactor overwrites the labs-map component from ember-mapbox-composer addon, and wraps it in a new basic-map component to separate external dependencies. Can you explain why both labs-map and basic-map mapboxGL abstractions are required?

This app was originally extending labs-map component javascript. I needed to override the related template because of a dependency upgrade to ember-mapbox-gl _in this project_. both e-mapbox-composer and this app _irresponsibly use private apis from ember-mapbox-gl_. So an upgrade to e-mapbox-gl led to some api changes, I needed to update this project to work with those changes correctly. 

The labs-map abstraction is required for this pull request because it includes untested behavior from before improvements to the test suite. BasicMap was added as an _initial_ step to introducing a layer  around e-mapbox-gl that we own. This PR tries to not refactor existing source code until the test suite is improved.

Issue #805, opened 2019-02-22
URL "root" query params get ignored, redirects to /about with defaults


allthesignals:
The URL "root" query params get ignored and the app redirects to /about with default params set.

For example, this URL:
https://zola.planning.nyc.gov/?layer-groups=%5B%22street-centerlines%22%2C%22tax-lots%22%5D#14.65/40.5933/-73.78866

Redirects into this:
https://zola.planning.nyc.gov/about?layer-groups=%5B%22building-footprints%22%2C%22commercial-overlays%22%2C%22street-centerlines%22%2C%22subway%22%2C%22tax-lots%22%2C%22zoning-districts%22%5D#9.72/40.7125/-73.733)

It also appears to ignore the lat/lon/zoom hash.

julialucyhogan:
this may be fixed by Chris's open PR #814

PR #851, opened 2019-06-07
Refactor main-map; fix routing bug with redirect

allthesignals:
Addresses #805

- This PR fixes a bug where direct, query-param'd links to main root of the app were redirecting to /about _without_ query params. Please review`services/layer-groups.js`
- Tests are added to demonstrate bug and increase coverage in main-map component
- As a rider, this PR splits out the measurement tools part of the map - specifically, that component is called map-measurement-tools. It shouldn't receive a thorough code review as it's largely moving code around. I would like input on the general direction (splitting it out into a component, the name of that component, using contextual params, etc).

### DIFF
--- a/app/services/layer-groups.js
+++ b/app/services/layer-groups.js
@@ -2,6 +2,14 @@ import Service from '@ember/service';
 import { A } from '@ember/array';
 import { copy } from 'ember-copy';
 
+const areIncomingLayerGroupsDefault = function(layerGroupsToShow, layerGroupsByDefault) {
+  const layersToShowAreDefault = layerGroupsToShow.every(
+    (layerGroupToShow, idx) => layerGroupToShow === layerGroupsByDefault[idx],
+  );
+
+  return layersToShowAreDefault;
+};
+
 /**
   Layer Group aggregate service
   Allows for computed properties on the aggregate state of layer groups.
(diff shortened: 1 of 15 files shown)


julialucyhogan:
Is this another case where you don't want to refactor the internals of the new component until there are tests, or should I be doing a real code review on the content of the component?

allthesignals:
> Is this another case where you don't want to refactor the internals of the new component until there are tests, or should I be doing a real code review on the content of the component?

Thanks, I'm avoiding a refactor of the internals of the new component until I can be more confident following more test coverage...

app/services/layer-groups.js
```diff
@@ -2,6 +2,16 @@ import Service from '@ember/service';
...
 import { copy } from 'ember-copy';
 
+const areIncomingLayerGroupsDefault = function(layerGroupsToShow, layerGroupsByDefault) {
+  const layersToShowAreDefault = layerGroupsToShow.every(
+    (layerGroupToShow, idx) => layerGroupToShow === layerGroupsByDefault[idx],
+  );
+
+  const currentDiffersFromDefault = layerGroupsToShow.length === layerGroupsByDefault.length;
+
+  return layersToShowAreDefault && currentDiffersFromDefault;
```
trbmcginnis: The naming here is great/intuitive, but I would love some comments explaining why this change was needed.

A little confused on why `currentDiffersFromDefault` would equal visible layer-groups having the same length as the default ones. Wouldn't that be the opposite?

Issue #798, opened 2019-02-11
Reset map layers button is broken


hannahkates:

trbmcginnis:
After this is fixed, put a test for this in `layer-palette-test.js`:

Something similar to:

  ```
test('clicking reset map button resets to all of default layers', async function(assert) {
    this.layerGroups = await this.owner
      .lookup('service:store')
      .findAll('layer-group', { include: 'layers' });

    this.layerGroupsObject = this.layerGroups.reduce(
      (accumulator, current) => {
        accumulator[current.get('id')] = current;
        return accumulator;
      },
      {},
    );

    this.selectedOverlays = [];
    this.selectedZoning = [];

    await render(hbs`
      {{layer-palette
        selectedZoning=selectedZoning
        selectedOverlays=selectedOverlays
        layerGroups=this.layerGroupsObject
        isDefault=isDefault
      }}
    `);

    await click('[data-test-grouped-parent="Commercial Districts"]'); // under Zoning Districts
    await click('[data-test-grouped-parent="C1-1 through C1-5"]'); // under Commercial Overlays
    await click('[data-test-toggle-aerials] .layer-group-toggle-label'); // Aerial Imagery
    const zoningDistricts = this.layerGroupsObject['zoning-districts'];
    const commercialOverlays = this.layerGroupsObject['commercial-overlays'];
    const aerialsLayer = this.layerGroupsObject.aerials;

    await click('[data-test-reset-map-button]');
    await pauseTest();
    assert.deepEqual(zoningDistricts.layers.findBy('id', 'zd-fill').style.filter, []); // reset to none
    assert.deepEqual(commercialOverlays.layers.findBy('id', 'co').style.filter, []); // reset to none
    assert.equal(aerialsLayer.visible, false); // toggle to false
  });
```

allthesignals:
@trbmcginnis where'd you land on this? Thanks for looking so closely into the problem.

I'm gonna try to dig in to the issue with the double render.

allthesignals:
Double render problem emanates from the layer group service

PR #852, opened 2019-06-07
849: add test for layer-palette

### LINKED ISSUE #849: Create test for layer-palette toggles
Write tests for layer palette toggling functionality. When a user toggles different layers, the changes should be reflected in the model.

trbmcginnis:
Note: I added a test (that is skipped for now) for the Reset Map button at the bottom of the layers palette. This button is broken right now (issue #798 ). This test can be used while fixing that bug in another branch.

Closes #849


allthesignals:
Beautiful!!!

Issue #847, opened 2019-06-06
Some lot views are failing 


allthesignals:
https://zola.planning.nyc.gov/lot/1/1331/7502?layer-groups=%5B%22building-footprints%22%2C%22commercial-overlays%22%2C%22street-centerlines%22%2C%22subway%22%2C%22tax-lots%22%2C%22zoning-districts%22%5D#17.58/40.760159/-73.964864

![image.png](https://images.zenhubusercontent.com/5ca23922d9fcf5278d715977/a933f9bf-226a-4a15-88c2-86709e43736b)

Related to network request... carto returns "query timeout exceeded"

allthesignals:
We are having a conversation with Carto to talk about upping the platform limits. We should handle request failures better in the run, however, but in the near term we'll see what Carto says...

PR #853, opened 2019-06-07
847 retry findRecord for lots; clean up serializer; remove blocking i…

allthesignals:
Addresses #847 by retrying the request for lots. The original problem was the Carto was responding with "query_limit_exceeded". This PR changes that so that it keeps retrying the request until it successful receives the data. Subsequent requests will continue until it receives a response that is not an error mentioning "query_limit_exceeded".

The main entry point to review is routes/lot, particularly the handling for errors. 

Some campsite rule riders include cleanup for the application serializer and removing an initializer that's not longer necessary.


allthesignals:
Closing until we talk to Carto cc @hannahkates

PR #857, opened 2019-06-10
846: remove unnecessary/separate out components in main-map

### LINKED ISSUE #846: Remove unnecessary components in main-map
Remove unnecessary/redundant components in main-map and separate out useful components.

trbmcginnis:
### Remove:

- **extra lot highlighting** functionality that is unnecessary, this was made obsolete with the implementation of `labs-layers-api`

- `currentAddress` which was being used for **placing a pin on addresses** back when we had a `map-search.hbs` component  ([commit history](https://github.com/NYCPlanning/labs-zola/commit/42e9e8b7a70eb4fa2fa43f9a6b1e000c6418d0fa#diff-7aa25e00ab4a18760720f946a551d98e) of `currentAddress`), now we just zoom to associated tax lot

- **extra layers** (under folder `app/layers`) that we are no longer using: `bookmarked-lots` we now define this layer in another component, `highlighted-lot` which was made obsolete by `labs-layers-api`, and `point-layer` which was the layer we were originally using to show the location of searched addresses (now we just locate the actual tax lots)

- `adjustBuildingsLayer` was an **unused action**

### Separate:

- **extra geolocate component** that is used for mobile is now in its own component `locate-me-mobile`

The only one left I can think of is `bookmarkedLotsLayer` (the outline on tax lots that have been bookmarked)--which can be placed in its own component. I am going to make a new issue for this. 

Closes #846


app/templates/components/main-map.hbs
```diff
@@ -29,7 +29,6 @@
     hoveredFeature=map.hoveredFeature
     layerGroups=this.layerGroups
     interactivity=interactivity
-    highlightedLineFeatureLayer=highlightedLotLayer
```
allthesignals: Are you sure about this? LabsMap lets you override the property: https://github.com/NYCPlanning/ember-mapbox-composer/blob/efca90484bb0a7e59ff95af640b4c180d93de62c/addon/components/labs-map.js#L95. 

Unless highlightedLotLayer is identical to what's in e-mapbox-composer.

allthesignals:
Nice but I would be sure that we're not taking away functionality

trbmcginnis:
> > currentAddress which was being used for placing a pin on addresses back when we had a map-search.hbs component (commit history of currentAddress), this was made obsolete by our search API
> 
> How does it add a geocoded address from search api to the map?

Sorry, not sure if I was totally clear. It looks like we originally had a component named `map-search` that handled our search functionality in the app. Along with that search component we had a point layer that would show up when a user searched an address, we used this computed `currentAddress` to check whether we should show this point layer:

```
  {{#if mainMap.currentAddress}}
    {{#map.source sourceId='currentAddress' options=mainMap.addressSource as |source|}}
      {{source.layer layer=mainMap.pointLayer}}
    {{/map.source}}
  {{/if}}
```

Now searching an address brings users to an entire tax lot rather than dropping a pin on the map.
This is what I gathered from when this code was originally implemented --> [Draw pin on map when selecting address commit](https://github.com/NYCPlanning/labs-zola/commit/42e9e8b7a70eb4fa2fa43f9a6b1e000c6418d0fa#diff-7aa25e00ab4a18760720f946a551d98e)

allthesignals:
@trbmcginnis gotcha, thanks, but do we not want the geocoded pin anymore? It does seem redundant if it's point to a lot. Just making sure.

trbmcginnis:
> @trbmcginnis gotcha, thanks, but do we not want the geocoded pin anymore? It does seem redundant if it's point to a lot. Just making sure.

Hmm, that's a good point, I guess it depends on whether there's ever an address that isn't associated with a tax lot which I feel like doesn't exist? We could keep the geocoded pin for a "just in case" situation where the address might not have a tax lot, but I feel like it'd be very rare

app/templates/components/main-map.hbs
trbmcginnis: Hmm okay I'll check on that

allthesignals:
Thanks just be sure to squash commits

app/templates/components/main-map.hbs
trbmcginnis: `highlightedLotLayer` highlights lots with a fill color of grey--I don't know when we would be using this, and the functionality for highlighting lots with a yellow outline comes from `ember-mapbox-composer`. So this also feels redundant, and I can't remember ever seeing another way to highlight lots in zola other than the yellow outline

app/templates/components/main-map.hbs
```diff
@@ -98,12 +91,5 @@
...
-{{!--
-  TODO: Geolocate Feature -> Move into separate component
---}}
-{{#unless findMeDismissed}}
-  <div class="find-me">
-    <button {{action 'locateMe'}} class="button hide-for-medium hide-for-print no-margin">Locate&nbsp;Me</button>
-    <button {{action 'dismissFindMe'}} class="button hide-for-medium hide-for-print no-margin">&times;</button>
-  </div>
-{{/unless}}
+{{#locate-me-mobile}}
```
julialucyhogan: why is this being invoked as a block component if nothing is invoked/rendered inside it?

app/templates/components/locate-me-mobile.hbs
```diff
@@ -0,0 +1,16 @@
+{{#unless findMeDismissed}}
+  <div class="find-me">
+    <button
+        onclick={{action 'locateMe'}}
```
julialucyhogan: nit but why are we using both "locate" and "find" as verbs here?

allthesignals: @julialucyhogan honestly, no idea... I looove the campsite rule, but I think we should be conservative about refactoring stuff in this app. Without rly good tests (which we lack), it almost always leads to weird regressions.

Stratgegy for us I think is to find the seams in the code, and split stuff up, but leave the code itself alone. Refactoring can come once this stuff is a little more testable.

Thoughts?

Issue #854, opened 2019-06-10
Clicking on zoning districts does not transition to appropriate route


trbmcginnis:
When a user clicks on a zoning district, there should be a transition to the zoning districts route. This is currently broken. 

To replicate bug, click on a zoning district and the zoning district info will not show up on the right panel as expected.

trbmcginnis:
First impressions:

- I think the error that we're seeing now: `Error while processing route: zoning-district Cannot read property 'search' of undefined TypeError: Cannot read property 'search' of undefined` have to do with the recent disabling of search for zoning districts. In this PR, 738: layer interactivity, https://github.com/NYCPlanning/labs-zola/pull/746, mixins were added that handled layer interactivity based on whether a user CLICKED on the layer or SEARCHED the layer (ex. searching for `R6A` zoning districts).

- When I remove this difference in layer interactivity of search vs. click in develop, clicking on the zoning districts transitions to the correct route (the panel on the right displays the correct information), but it is going very slowly and there are often

PR #859, opened 2019-06-10
854: QuickFix, remove search as an option from mixin while search is deactivated

trbmcginnis:
While search is deactivated for zoning districts, commercial overlays, zoning map amendments, special purpose districts, and special purpose subdistricts, this PR removes the `search` option from a Mixin that determines search vs. click layer interactivity. 

The deactivation of search for these layers was causing clicking on a layer to not transition to the appropriate route--this should remove that error for now. 

This is a quick fix and will be reverted once search is working again. 

Addresses #854

### DIFF
--- a/app/mixins/update-selection-all-features.js
+++ b/app/mixins/update-selection-all-features.js
@@ -11,7 +11,7 @@ import { inject as service } from '@ember/service';
 export default Mixin.create({
   mainMap: service(),
   afterModel({ taskInstance }, transition) {
-    if (transition.queryParams.search === 'true') {
+    if (transition.queryParams && transition.queryParams.search === 'true') {
       this.waitToFitBounds.perform(taskInstance);
     }
   },


app/mixins/update-selection-all-features.js
```diff
@@ -10,11 +10,12 @@ import { inject as service } from '@ember/service';
 
 export default Mixin.create({
   mainMap: service(),
-  afterModel({ taskInstance }, transition) {
-    if (transition.queryParams.search === 'true') {
-      this.waitToFitBounds.perform(taskInstance);
-    }
-  },
```
allthesignals: This all seems a little heavy-handed, no? Why don't you just add a defensive check on line 13? Then we don't have to comment out your enhancement :)

```
if (transition.queryParams && transition.queryParams.search === 'true') {
  // stuff
}
```

trbmcginnis: Oh yes much better--I'll make that change

trbmcginnis: Done!

Issue #806, opened 2019-02-22
Encoding/decoding url param array unstable


allthesignals:
Encoding/decoding url params array is unstable. I think this happens when links are pasted across context, causing it to "rot" in a way through accidental re-encoding of encoded URLs.

The solution is to use the more "native" URL param array list:

`layer-groups[]="something"&layer-groups[]="another thing"`

This is more stable and will come parsed as an array. No manual parsing necessary.

allthesignals:
This seems to happen when encoded URLs get hyperlinked from a document. Then the browser will _re-_encode the URL:

1. %3Flayer-groups%3D%5B%22street-centerlines%22%2C%22tax-lots%22%5D%2314.65%2F40.5933%2F-73.78866
2. %253Flayer-groups%253D%255B%2522street-centerlines%2522%252C%2522tax-lots%2522%255D%252314.65%252F40.5933%252F-73.78866

This double encoding problem may be unavoidable, will need to keep researching.

PR #860, opened 2019-06-11
Use router service; find more seams, extract;

allthesignals:
# What
This PR addresses #806 by refactoring the LayerGroups service class to make it clearer where decoding/encoding of URL params is and isn't happening.

It isn't clear where the _original_ issue is coming from, whether it's something we can address given the sensitivity to the URL routing. This isn't really a fix as much as it's clean up, bringing clarity to the problem...

# How 
 - Refactor the layer-groups service.

# Entrypoint 
 - see app/services/layer-groups.js

# Riders
 - I extracted some functionality out of main-map into separate components. Their behaviors are already tested in acceptance tests.


app/services/layer-groups.js
```diff
@@ -1,107 +1,136 @@
...
+export const isEqual = function(array1, array2) {
+  array1.sort();
+  array2.sort();
 
-  return layersToShowAreDefault;
+  return array1.every((value, index) => value === array2[index]);
+};
+
+const matchesIdentifierFor = function(layerGroup) {
+  return param => (param.id || param) === layerGroup.id;
```
godfreyyeung: Where is param coming from?

app/services/layer-groups.js
```diff
@@ -1,107 +1,136 @@
...
+};
+
+export const applyQueryParamToLayerGroup = function(layerGroup, layerGroupIDsToShow) {
+  const paramForLayerGroup = layerGroupIDsToShow
+    .find(matchesIdentifierFor(layerGroup));
+
+  set(layerGroup, 'visible', !!paramForLayerGroup);
+
+  // handle layer group substate
+  if (paramForLayerGroup && paramForLayerGroup.selected) {
```
godfreyyeung: If `.selected` is a boolean, then this block is equivalent to 
```
 if (paramForLayerGroup && paramForLayerGroup.selected) {
    set(layerGroup, 'selected', *true*);
}

```
Just noting it here in case it helps debugging down the road.

app/services/layer-groups.js
allthesignals: Oy! It's confusing. it's the param to an array function (which has no parentheses). This func returns a function - (is this "currying"?).

allthesignals: In this case it is passe to an array method, param comes from typical iteratable

app/services/layer-groups.js
allthesignals: Thanks @godfreyyeung, I actually don't know for sure haha. I believe it is a string, however - it's a specific layer-id.

Issue #268, opened 2017-10-06
[Refactor] Abstract routable layers


allthesignals:
We need multiple routes for layers, but there's duplication across these routes, including controllers, route logic, and so on. Layer-specific templating should be moved into components and can be identified through a generic route, relying on url tokens.

allthesignals:
The problem is that any layer that should behave differently will need its own route. Need to research more.

hannahkates:
@allthesignals Is this still relevant to your more recent URL refactoring plans or can it be closed?

PR #861, opened 2019-06-12
Add generic route for “second class” layer group views;

allthesignals:
This PR addresses #268 by adding a generic route for "second class" layer group views - things like zoning map amendments, which have their own template logic but essentially have the same routing logic in terms of the route's dynamic segments.

I think this helps with maintainability a bit more by remove duplication of routes and leveraging components instead.

Most "additions" in the diff come from tests generated for components used to display the modified layer group views.


app/components/layer-group-display/zoning-district.js
```diff
@@ -0,0 +1,29 @@
...
+  @service
+  mainMap;
+
+  @computed('model.value.primaryzone')
+  get primaryzoneURL() {
+    const primaryzone = this.get('model.value.primaryzone');
+    let url = '';
+
+    if ((primaryzone === 'c1') || (primaryzone === 'c2')) {
+      url = 'c1-c2';
```
julialucyhogan: feels weird to call this string a "URL"

allthesignals:
Closing in favor of another PR that splits this out much more

PR #866, opened 2019-06-17
Add failing behavior grid tests

allthesignals:
This PR adds a much-needed implementation of tests based on the "layer behavior grid". This test suite will let us move move forward with code improvements.

Note that there are 5 of these tests that are set up as "todo" tests, meaning they fail (in 1 of 8 assertions), but may still be merged as they are behaviors that do not match up with what the behavior grid asks for.

Also, Percy snapshots don't work for the map just yet, and I don't have any plans to do it. Getting Percy to work with the map requires doing additional complex test-environment-specific work. I'd like to find a more general solution.

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -19,7 +19,7 @@ if (typeof Ember.Test === 'undefined') {
   Ember.Test = {
     registerWaiter() {
 
-    }
+    },
   };
 }
 /* eslint-enable ember-suave/no-direct-property-access, no-undef */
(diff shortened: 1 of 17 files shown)


godfreyyeung:
Is there a link or reference to the "layer behavior grid"?

allthesignals:
> Is there a link or reference to the "layer behavior grid"?

Yes! Forgot to link: https://airtable.com/tbl7qrEzVgOlHsqBR/viw85aFEf3ld7KFa2

PR #868, opened 2019-06-19
268 refactor route abstractable

### LINKED ISSUE #268: [Refactor] Abstract routable layers

allthesignals:
Hi!

This pull request addresses #268 ...

This is obviously in "draft" form because I want to get some preliminary feedback from the team on meta issues, specifically, "is this reviewable in spite of how large it is?" Additionally, I made this draft form so that I could visualize better what the diffs would look like as well as to see the Percy diffs run.

While my tests pass, there are a few visual regressions in the map-feature detail views which I should investigate. 

@hannahkates please take a look at the Percy diffs - reminder that the map will be blank for now. I'm mostly looking for documentation of visual issues (should be moderate/severe issues, ie missing information). I notice that some things appear off by a few pixels. I don't think pixel-perfect is worth our time right now and would argue that _visual diffs by the magnitude of a few pixels should be ignored_. The lot view does appear to need a lot of improvements, but zoning districts views are downshifted a bit, for example.

Also, @hannahkates, in terms of the deploy preview, I think we should look through it and manual test / sanity check things by comparing to production. I'm interested in discrepancies between production and this branch, not bugs or regressions that exist in both branches. To make sure that existing links work, we should pull a list and add them to the test suite. Google Analytics will be a trove of stuff! But also random e-mails we've received.

@NYCPlanning/engineering here's an overview:

This pull request makes 2 major changes:
 - **Generalizes most of the behavior for "map-feature" detail routes.**
    Map feature detail routes ("lots/1", "zoning-district/1") were confusing, used multiple mixins, and had a lot of duplicated code. This pull request helps address that by centralizing all the map-feature-specific "presenter" logic into map-feature-specific components.

    This means, for example, that any text-manipulation required for human-friendly consumption of data (BBL should be split, add dashes into zoning districts, etc), will be owned by that map-feature's own _component_. 

    - Map feature behavior – including rendering and bounds-fitting – is [now encapsulated](https://github.com/NYCPlanning/labs-zola/pull/868/files#diff-2854319a1a2f31e1dade76f8d9c49fb6) in a component invoked by the map-feature's own route. Originally this was spread out across two mixins (TODO: delete these mixins). Additionally, these behaviors were triggered implicitly by route event hooks. The new component implementation is a little more explicit and declarative, rather than relying on the sometimes-chaotic internals of routes.

 - **Refactors map-feature models and bookmarks**
    Bookmarks were confusing (and still are), but now they are a little more [centralized](https://github.com/NYCPlanning/labs-zola/pull/868/files#diff-84edeae4a8929a052dd991b14d45f584) inside a bookmarking component. Originally, bookmark record-creation was handled by map-feature-specific controller actions. Because controllers for these routes were removed, these actions are managed in one place.

    Additionally, this adds a 'carto-geojson-feature' model, which extends Bookmark. Because they are GeoJSON features pulled from Carto, map-features themselves now extend CartoGeojsonFeature. This adds a little more consistency when interacting with these features. It also makes it easier to represent information in the bookmarks views because they add a consistent `properties` attribute.

Tests are beefed up a little bit, with a large "layer group behavior test" already merged into develop for the purpose of comparison... you will notice that Codecov is not happy. This is partly due to a few things: 

First, "hands-off" refactoring being more concerned with how groups of behavior are arranged and interact with each other rather than the internals of those groups. To codecov, this will look like a bunch of new untested file additions, hence a really low score.

Second, there are some old files I need to remove...

@NYCPlanning/engineering I'd say the main entry points would be `router.js`, `routes/map-feature`, and the subroute templates you find under `map-feature`.

Let me know how I can make this clearer and if I should reorganize these commits. Tests are broken when I try to divide this commit history further, so I'd need to go back and re-fix tests throughout time, which might be a little arduous...


app/models/commercial-overlay.js
```diff
@@ -1,21 +1,3 @@
-import DS from 'ember-data';
-import { computed } from '@ember/object';
-import bbox from '@turf/bbox';
-import Bookmarkable from './bookmark';
+import CartoGeojsonFeature from './carto-geojson-feature';
```
julialucyhogan: <3 this

PR #924, opened 2019-07-15
642 restore zoom dependent warning labels

godfreyyeung:
## What 

Restores zoom dependent warnings.

## How

Takes advantage of @andycochran 's `activeTooltip` parameter on the `layer-group-toggle` component. 

Created a new helper to either return a null value or the warning label to `activeTooltip`. For a given layer group, the helper returns null when the map zoom is higher than all sublayers' minzoom, otherwise it returns a message about possible missing information.  

## Riders
Detached the `main-map` service's `zoom` property from the app url hash.

### DIFF
--- a/app/templates/components/layer-palette.hbs
+++ b/app/templates/components/layer-palette.hbs
@@ -17,6 +17,7 @@
     <container.layer-group-toggle
       data-test-toggle-tax-lots
       @layerGroup={{this.layerGroups.tax-lots}}
+      @activeTooltip={{zoom-dependent-label this.layerGroups.tax-lots this.zoomWarningLabel this.mainMap.zoom}}
     >
       <label
         class="checkbox-wrapper"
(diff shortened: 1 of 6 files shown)


app/templates/components/layer-palette.hbs
```diff
@@ -17,6 +17,7 @@
     <container.layer-group-toggle
       data-test-toggle-tax-lots
       @layerGroup={{this.layerGroups.tax-lots}}
+      @activeTooltip={{zoom-dependent-label this.layerGroups.tax-lots this.zoomWarningLabel this.mainMap.zoom}}
```
allthesignals: This is great!

Do you think it would be possible to set this up in the contextual component that is yielded from the layer group container? See here: https://github.com/NYCPlanning/labs-zola/blob/develop/app/templates/components/labs-ui/layer-groups-container.hbs#L8-L15

That way, there's way less copy pasta... unless there's a _lot_ of variety across these components. I think it would save some overhead. You could definitely pass the other dependent arguments into the container.

godfreyyeung: Ohhh good idea. Yeah I definitely thought about yielding it from `layer-group-container` since I thought it should be a sibling component/helper of `layer-group-toggle`, instead of wrapping it or being nested in it. I just didn't think of using one sibling component as a "helper" for another! I can give refactoring a go. 

I'm also starting to think maybe `layer-group-toggle` could have a parameter for a condition or function condition for showing the tooltip. I am cautious about expanding the `layer-group-toggle` api too much, though

Issue #484, opened 2018-01-03
Poor Firefox performance


allthesignals:
User has reported strange behavior, specifically poor performance in Firefox. This can be observed on different machines. 

Visiting this link: https://zola.planning.nyc.gov/lot/5/924/1?zoning-districts=false#13.86/40.5972/-74.1039

Details:
 - Browser: Firefox
 - Browser version: 53.0 (32-bit)
 - OS: Windows 7 Pro, SP1

Full report:
> Selecting and deselecting parcels in and around Henry Kaufmann Campgrounds on SI, the screen would just grey out. Specifically, the map and information pane on the right would go blank. Sometimes this would happen without selecting/deselecting. I would zoom in, sometimes the zoom itself would be too much and the map would go blank. Refresh, zoom in, maybe select 1 lot. If I tried to deselect, the information pane and the map would go blank.
>
>To find what tax lots made up this zoning lot (it was a little over a dozen), I had to refresh a bit more than 2 dozen times.

allthesignals:
I'm going to look into tackling this with some general performance improvements...

allthesignals:
We've made a lot improvements to performance, closing for now

PR #935, opened 2019-07-17
934 prember

### LINKED ISSUE #934: Add prember for some initial views
A compromise with full fastboot is to use https://github.com/ef4/prember which will pre-render some views at build time. Not good for dynamic content but works fine for static views.

allthesignals:
This PR enables build-time pre-rendering to improve the initial load time for the index (/about) page. It shaves approximately 2 seconds from "first contentful paint", bringing the contentful paint down to 1 second. This addresses #484 by improving the perception of speed.

Closes #934.

This pull requests adds ember fastboot and "prember". 

A lot of research went into this — there are numerous challenges with getting fastboot to work _well_, especially with an app like ZoLa, which has over a million possible routes. Running a live fastboot server for ZoLa would introduces large maintenance overhead. We would need to setup a caching layer like Varnish to speed up generated routes. We would also need to spend a lot more time researching, understanding and addressing some large performance issues and memory leaks occurring the app now, which causes major memory overloads on a fastboot production server.

This PR is more of a compromise — we get _some_ pre-rendering for the initial routes, but not for dynamic routes (lot views), and we _don't_ have to deal with the maintenance overhead of running a live server. Should this become an interest later on down the road, I think we at least have this experience to go off of.

Additional changes:
 - Change adapter inheritance to extend from one adapter class, defined in app/adapters/application.
 - Adds a number of "if-guards" for fastboot — because node is pre-rendering some views _at build time_, application code is being run in a different environment that does not have access to things like `window` and `document` globals. These if-guards exist to enable fastboot.
 - Removes some array extensions like `mapBy` and `filterBy`, which are not supported in the node-based ember compiler.
 - Netlify redirects are adjusted to support statically-generated routes, but default to dynamic routes
 - config/environment is adjusted to handle API hosts differently. Because the `--environment=production` or `staging` flag has special meaning during the ember build step, we should use a different approach to configuring these endpoints for staging.

Screen Shot 2019-07-17 at 9 15 07 PM


allthesignals:
@julialucyhogan 

`!fastboot.isFastBoot` <- the code is run on the browser
`fastboot.isFastBoot` <- the code is run in node

PR #945, opened 2019-07-22
Refactor & style AnnouncementBox

andycochran:
- Moves the AnnouncementBox component to the site header so it's available on all routes
- Uses Ember Tooltips instead of custom action to toggle the announcement (added benefit of clicking anywhere on page to close it)
- Adds `what-input` to prevent `:focus` outline styles showing for non-keyboard navigation
- Adds some temporary overrides for Labs UI and Ember Tooltips (should be remove once [this PR](https://github.com/sir-dunxalot/ember-tooltips/pull/370) has been merged/published)
- Updates the language of the announcement to be more friendly/positive

### DIFF
--- a/app/templates/components/announcement-box.hbs
+++ b/app/templates/components/announcement-box.hbs
@@ -1,14 +1,18 @@
-<div
-  class="button expanded tiny text-left no-margin"
-  style="text-align: left;"
-  onclick={{action this.toggle}}
->
-  {{fa-icon 'bullhorn'}}
-  Announcement: {{this.title}}
-</div>
-
-{{#if this.isOpen}}
-  <div class="announcement-box" style="background-color: #ecedee">
-    {{yield}}
-  </div>
-{{/if}}
+<button>
+  {{fa-icon "bell"}} <sup class="a11y-orange">NEW!</sup>
+  <EmberPopover
+    @side="bottom-end"
+    @event='click'
+    @tooltipClassName='ember-popover dark'
+    @popperContainer='body'
+    as |popover|
+  >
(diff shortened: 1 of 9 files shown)


app/templates/components/announcement-box.hbs
```diff
@@ -1,14 +1,18 @@
...
+    @popperContainer='body'
+    as |popover|
+  >
+    <p class="text-small">
+      We've been working on ZoLa to prevent service interruptions. With these upgrades, you should now notice performance improvement.
+    </p>
+    <p class="text-small">
+      Thank you for using ZoLa!
+      <br>—<a href="mailto:zola_gis@planning.nyc.gov"> NYC Planning Labs</a>
+    </p>
```
allthesignals: Looks great! Why keep this inside the component instead of passing it as a block from the invoking side?

allthesignals:
@andycochran 

> Adds what-input to prevent :focus outline styles showing for non-keyboard navigation

Can we address this in a separate isue?

app/templates/components/announcement-box.hbs
andycochran: That would work. I had the whole thing in the site header, without using a component at all. I hastily moved it into the component. Makes sense to use it as a block component with the content in the header.

andycochran:
The reason I added what-input into this PR instead of addressing separately is because the new item  in the site menu is a `<button>` element (the others are `<a>`) and when you click it you get the focus style. 

image

You can notice the focus style on other elements too (like the dropdown in the BBL Lookup).

PR #966, opened 2020-01-03
965: add inwood disclaimer on lot, special purpose, and zoning map am…

### LINKED ISSUE #965: Add disclaimers for Inwood rezoning
**Places to add disclaimer:**
1. Lot profile
2. Zoning map amendments
3. Special purpose district

[**Example disclaimer to copy**](https://github.com/NYCPlanning/labs-zap-search/blob/048d58f8a71f8f63fcecfdddbae88def0c796a96/app/templates/my-projects/assignment/hearing/done.hbs#L26-L28)

**Disclaimer text:**
"Note: All zoning regulations pertaining to the Special Inwood District Rezoning, which amended the Zoning Resolution (N 180205A ZRM) and Zoning Map (C 180204A ZMM), and which have been in effect since 8/8/18, are no longer in effect as of 12/19/19 per court order. For the applicable zoning designations currently in effect, please see zoning maps 1b, 1d, 3a, and 3c."

1b link: https://www1.nyc.gov/assets/planning/download/pdf/zoning/zoning-maps/map1b.pdf
1d link: https://www1.nyc.gov/assets/planning/download/pdf/zoning/zoning-maps/map1d.pdf
3a link: https://www1.nyc.gov/assets/planning/download/pdf/zoning/zoning-maps/map3a.pdf
3c link: https://www1.nyc.gov/assets/planning/download/pdf/zoning/zoning-maps/map3c.pdf

hannahkates:
Adds the Inwood rezoning disclaimer on lot, special purpose, and zoning map amendment profiles.

For the tax lot profiles, the display of the disclaimer is dependent on a new field being added to mappluto called `notes` which will contain number codes. The number for the Inwood rezoning lawsuit contention will be `1` stored as a numeric. To test these code changes, I added a `notes` column to the table on Carto and set notes=1 for two different BBLs: 1022250010, 1022380023. 

EDM will soon provide updated mappluto data with all the appropriate lots tagged with 1 in the notes field. We shouldn't merge this work into master until we have that updated data.

Closes #965

**Screenshots:**

**Tax Lot**
image

**Special Purpose District**
image

**Zoning Map Amendment** 
image

### DIFF
--- a/app/templates/components/layer-record-views/tax-lot.hbs
+++ b/app/templates/components/layer-record-views/tax-lot.hbs
@@ -210,6 +210,11 @@
       </ul>
     </div>
   </div>
+  {{#if (eq this.model.notes 1)}}
+    <div class="callout secondary">
+      <p>{{fa-icon 'exclamation-triangle' class='orange-muted'}} Note: This lot is within the Inwood Rezoning area. All zoning regulations pertaining to the Special Inwood District Rezoning, which amended the Zoning Resolution (N 180205A ZRM) and Zoning Map (C 180204A ZMM), and which have been in effect since 8/8/18, are no longer in effect as of 12/19/19 per court order. For the applicable zoning designations currently in effect, please see zoning maps <a href="https://www1.nyc.gov/assets/planning/download/pdf/zoning/zoning-maps/map1b.pdf" target="_blank">1b</a>, <a href="https://www1.nyc.gov/assets/planning/download/pdf/zoning/zoning-maps/map1d.pdf" target="_blank">1d</a>, <a href="https://www1.nyc.gov/assets/planning/download/pdf/zoning/zoning-maps/map3a.pdf" target="_blank">3a</a>, and <a href="https://www1.nyc.gov/assets/planning/download/pdf/zoning/zoning-maps/map3c.pdf" target="_blank">3c</a>.</p>
+    </div>
+  {{/if}}
 </section>
 <hr class="hide-for-print" />
 <section class="lot-details">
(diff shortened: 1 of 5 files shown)


app/templates/components/layer-record-views/tax-lot.hbs
```diff
@@ -210,6 +210,11 @@
       </ul>
     </div>
   </div>
+  {{#if (eq this.model.notes 1)}}
```
allthesignals: ~Can we do `{{#if this.model.notes}}`~

Nevermind, I see that the numeric from EDM's change to PLUTO corresponds to a specific piece of text in the template. The idea is that there could be more notes, and you might want to show them all.

One problem I'm seeing is that there is duplicate text spread across three files. I'd really like to see a single component used for these callouts.

andycochran: Right. The `notes` field _could_ be a comma separated list — e.g. `1,2,4`. 

Long-term solution, would be a notes component that's passed an array and loops thru them to output multiple `div.callout` elements. Each actual note content will ultimately be provided in a separate table, so there'd need to be a query for those. Tho for now, I think it's okay the the note content is hardcoded, but agree that it should only be hardcoded once. 

On the other hand, this exact situation being solved for is very much an edge case, has not happened before, and will likely not happen again. So maybe it's fine we don't overthink it… LGTM.

allthesignals: Yes that is one way to do it. If you need help doing it let me know or bounce this over to me please.

hannahkates: The text on the tax lot view is slightly different actually. Ideally we will be undoing this work in a few months once the lawsuit is resolved. If future note demands come up, I think we should deal with componentizing and improving it then, but I'll defer to @allthesignals if he wants to implement it differently.

trbmcginnis:
this is matt

PR #969, opened 2020-01-23
962: only show cyclomedia button, not embed

### LINKED ISSUE #962: Provide cyclomedia link instead of embedding
The cyclomedia embed slows down the app load time. Most users don't scroll down far enough to see the cyclomedia window, so we should get rid of the embed and just provide a hyperlink instead. That will improve app performance, especially on mobile.

hannahkates:
Closes #962. Gets rid of conditional logic for providing cyclomedia access as an embedded widget vs. a button that opens in a new tab. Now we only provide the button.

### DIFF
--- a/app/templates/components/road-view.hbs
+++ b/app/templates/components/road-view.hbs
@@ -1,18 +1,9 @@
 <div class="cyclomedia hide-for-print">
-  {{#if (and this.inView (media "desktop"))}}
-    <iframe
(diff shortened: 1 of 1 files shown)


allthesignals:
Thank you!

What kind of performance problems were you observing? The Cyclomedia iFrame only loads if you scroll down far enough. Just want to make sure there isn't something else going on.

PR #990, opened 2020-08-10
reCAPTCHA owners

andycochran:
This PR adds reCAPTCHA to owner field on tax lots.


andycochran:
Validation is encapsulated in the widget. The `reCaptchaResponse` token is proof that the user completed the bot check. However, it does not prevent HTTP client request forgery. That's not a problem for this app. But we could validate the token with a `POST` to `https://www.google.com/recaptcha/api/siteverify?secret=[secretkey]&response=[token]` and check the response: `{ "success": [true|false], ... }`

andycochran:
Tests/devops have been trouble for me here. 😭

- The `ember-g-recaptcha` addon requires node 10, so I've had to drop support for node 8
- In all the ember tests that load the info panel (`map-feature.hbs`), I had to add a `waitFor` to make sure `ember-g-recaptcha` was done doing its thing, or it borks the test when it's destroyed. 
- I managed to get it to deploy on Netlify by adding a `.node-version` file (Netlify locks your site to the node version you're already using, so you have to explicitly change it)
- CI still seems to not care for the changes, but I'm not sure why

I was hoping not to disrupt the team, but might need to hand this off…

PR #995, opened 2020-08-25
update sitemap files

andycochran:
This PR… 
- Runs `scripts/sitemap.js` to regenerate the sitemap text files
- Updates `robots.txt` with the new sitemap files, and disallows `/bookmarks`

(We should automate this on production deploys.)


hannahkates:
Ideally this should get updated each time new MapPLUTO is published, not just when ZoLa code changes are deployed.

PR #1014, opened 2021-05-04
2492 update readme

dhochbaum-dcp:
This PR 
- Adds font-awesome dependencies to resolve a build error
- Adds pre-commit testing and linting
- Updates the readme

Closes [AB#2492](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/2492)


package.json
```diff
@@ -138,5 +143,8 @@
   },
   "fastbootWhitelist": [
     "abortcontroller-polyfill"
-  ]
+  ],
+  "lint-staged": {
+    "*.js": "eslint --cache --fix"
```
allthesignals, line 148: Thank you!

.husky/pre-commit
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npm test
```
allthesignals, line 4: I could be wrong but both `npm` and `npx` could be changed to `yarn run`.

I would make these both `yarn` because that's what most of our projects use.

allthesignals (approved):
Just one comment on the package manager command used (`yarn` over `npm`/`npx`) and lastly:

Could you try to make a few local junk commits to make sure we don't have anything funky going on with the pre-commit hook?

PR #1016, opened 2021-05-25
Add new tooltip language for special purpose subdistricts

rtblair:
Fixes [AB#2940](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/2940).

### DIFF
--- a/mirage/static-fixtures/layer-groups.js
+++ b/mirage/static-fixtures/layer-groups.js
@@ -263,7 +263,8 @@ export default {
         visible: false,
         legend: {
           label: 'Special Purpose Subdistricts',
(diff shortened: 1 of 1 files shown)


allthesignals (approved):
I think you want https://github.com/NYCPlanning/labs-layers-api (unless I totally don't get it)

I think we centralized tooltip copy in labs-layers so we could couple it with layer sources.

The files you've edited here are "fixtures" used to run automated tests (with out the need for external dependencies over HTTP).

rtblair:
@allthesignals Got it. Will do. Closing.

PR #1020, opened 2021-07-07
Sync dev with hotfixes on master

rtblair:


TylerMatteo (approved):
I'm going to approve and merge this. The linting problems will get fixed by a follow up PR into develop that I'm working on.

PR #1038, opened 2021-12-06
Updating labs-ember-search to version with correct sql query and fixing linting violations

TylerMatteo:
This PR updates the version of labs-ember-search to a new version that has corrected carto data set names and fixes a bunch of linting violations that found their way into master (and then develop) via some hotfix commits.

### DIFF
--- a/app/adapters/special-purpose-district.js
+++ b/app/adapters/special-purpose-district.js
@@ -2,7 +2,7 @@ import { buildSqlUrl } from '../utils/carto';
 import CartoGeojsonFeatureAdapter from './carto-geojson-feature';
 
 const SQL = function(id) {
-    return `SELECT cartodb_id as id, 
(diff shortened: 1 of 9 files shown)


godfreyyeung (approved):

app/adapters/special-purpose-district.js
```diff
@@ -12,10 +12,10 @@ const SQL = function(id) {
 };
```
rtblair, line 12: Note that we should be using `dcp_special_purpose_districts` here.

TylerMatteo: Yea I'll capture that in a bug or task. It also looks like `zoning-map-amendment.js` references `zoning_map_amendments` but a data set called `dcp_zoning_map_amendments` exists.

PR #1043, opened 2021-12-22
6068 Some lots are missing special purpose districts

dhochbaum-dcp:
Updated to proper carto table, renamed env variables to match carto data.

Closes #AB6068


godfreyyeung (approved):
I don't have a lot of context but looks good in terms of code

PR #1056, opened 2022-06-14
8269 'Special Purpose Subdistricts' data does not show

dhochbaum-dcp:
This removes an variable "subsub" that does not exist in the current version of the subdistricts table.  Previous version of the table did contain the column, but all values were NULL.

Closes #AB8269


horatiorosa (approved):
✅

PR #1057, opened 2022-06-14
7514 Zola Issue: Special Districts (2) property search results not displaying info for newly added districts

dhochbaum-dcp:
This PR adds Gowanus to the environment file.  Bug also mentions Jerome Ave, which had been fixed in a previous update.

Closes #AB7514


horatiorosa (approved):
✅

PR #1066, opened 2022-08-29
10276 Zola GA4 Implementation

dhochbaum-dcp:
This PR implements Google Analytics 4 tagging.

Only known issue is 'download_bookmark' event is being double-counted.  I have investigated and have no explanation for why it is occurring.

Closes #AB10276 and #AB9462

### DIFF
--- a/app/utils/gtag.js
+++ b/app/utils/gtag.js
@@ -1,5 +1,5 @@
-export default function gtag(...args) {
+export default function gtag() {
   window.dataLayer = window.dataLayer || [];
-  window.dataLayer.push(args);
-  return args;
+  const event = arguments // eslint-disable-line
+  window.dataLayer.push(event);
 }
--- a/app/utils/gtag.js
+++ b/app/utils/gtag.js
@@ -1,0 +1,5 @@
+export default function gtag(...args) {
+  window.dataLayer = window.dataLayer || [];
+  window.dataLayer.push(args);
+  return args;
+}
--- a/app/components/bookmarks/bookmark-button.js
+++ b/app/components/bookmarks/bookmark-button.js
@@ -1,6 +1,7 @@
 import Component from '@ember/component';
 import { action, computed } from '@ember/object';
 import { inject as service } from '@ember/service';
+import gtag from 'labs-zola/utils/gtag';
 
 export default class BookmarkButton extends Component {
   bookmarkableModel = null;
@@ -24,6 +25,11 @@ export default class BookmarkButton extends Component {
     const resolvedBookmark = await bookmark;
 
     if (resolvedBookmark) {
+      gtag('event', 'delete_bookmark', {
+        event_category: 'Bookmark',
+        event_action: 'Deleted Bookmark',
+      });
+
       resolvedBookmark.deleteRecord();
       resolvedBookmark.save();
     } else {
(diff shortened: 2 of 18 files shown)


app/utils/gtag.js
```diff
@@ -0,0 +1,5 @@
+export default function gtag() {
```
TylerMatteo, line 1: So why doesn't this function take any arguments? It looks like you're calling it with arguments when you use it.

dhochbaum-dcp: I tried having the function take in arguments and then passed that object on to the dataLayer.  Unfortunately, it didn't seem to pass the event on to GA when I created the object, but worked fine when I passed along the Arguments object.  I wish I had a better explanation, but I don't.

TylerMatteo: Can you point me to any documentation that you based this bit of code off of? I'm only asking because I was under the impression that the `<script>` tag you added to index.html loads a global function called `gtag` that you can call directly to track events. If that function exists then you might be overwriting here which could explain some of the weird behavior with the params.

TylerMatteo (changes requested):

TylerMatteo:
@dhochbaum-dcp Do these changes always log these events regardless of the environment the app is running in? I think we should at least be able to turn it off when running the app locally by default (with an ability to enable it if you're working on the events themselves). I can see the value of have them enabled in staging for testing but would we able to filter events by environment in the GA4 UI?

app/utils/gtag.js
dhochbaum-dcp: I can't point to any documentation I based this code off of.  I can't find any documentation on how to set up GA4 on Ember.

You are correct, the script added to index.html does add a global function called `gtag` that you can call directly to track events.  Unfortunately, without declaring gtag as a function within our code, none of our tests pass because we are calling an undeclared function.  We are definitely overwriting here, I just don't know of a better solution that both works and passes tests.

dhochbaum-dcp:
> 
These changes will log the events regardless of the environment.  We are able to filter by hostname in GA4, so we can remove any traffic that is not on production.  That's not the ideal way, I would much prefer to have it disabled by default in development, but we can go ahead and push this change now and I can add implementing that to my to-do list as a separate task.

app/utils/gtag.js
TylerMatteo: Understood. I was referring to just documentation for gtag.js in general but I see that what you have up in index.html follows their docs. I'd like you to spend some more time trying to get the tests to pass without the need for this util. Having competing version of this function is likely to cause us headaches and we should be able to mock this in the tests somehow without having a whole function that only exists to make our tests happy.

Shot in the dark here but don't suppose you've tried adding your snippet in index.html to the index.html located in the `tests` folder?

dhochbaum-dcp: Adding the snippet to the tests/index.html did not appear to have any effect on the outcome of the tests.

I was able to get tests to pass without the additional function by excluding 'no-undef' from .eslintrc.js, and initial tests showed that it was tracking correctly in GA.  If you would prefer that approach, I can test and make sure all of the tracking is correct and then open a new pr.

dhochbaum-dcp:
Closing, replaced by #1067

PR #1068, opened 2022-09-14
updated api request to use  borocode column values

horatiorosa:
#### Updates:
  - `app/adapters/lot.js`to add `borocode` to the `LotColumnsSQL` API request.
  - `app/models/map-features/lot.js` to use `map_pluto`'s `borocode` instead of the first digit of the `cd` value (in this case, it was incorrectly using 2 for the borough code instead of 1).
 - `app/templates/components/layer-record-views/tax-lot.hbs` to use new getter with correct column values
 - `tests/unit/models/lot-test.js` to include `borocode` to fix failing test
 -  updated failing acceptance tests to include a `properties` object with `borocode`

 
### Closes:
 [AB#10530](https://dev.azure.com/NYCPlanning/ITD/_workitems/edit/10530)

#### Before:
WRONG

#### After:
RIGHT

### DIFF
--- a/app/models/map-features/lot.js
+++ b/app/models/map-features/lot.js
@@ -323,6 +323,8 @@ export default class LotFragment extends MF.Fragment {
 
     @attr('string') bldgclass;
 
+    @attr('string') borocode;
+
     @attr('number') lat;
 
     @attr('number') lon;
@@ -412,12 +414,6 @@ export default class LotFragment extends MF.Fragment {
       return bldgclassLookup[this.bldgclass];
     }
 
-    @computed('cd')
-    get borocode() {
-      const borocd = this.cd;
-      return `${borocd}`.substring(0, 1);
-    }
-
     @computed('borough')
     get boroname() {
       return boroughLookup[this.borough];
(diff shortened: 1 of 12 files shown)


app/models/map-features/lot.js
```diff
@@ -412,10 +414,10 @@ export default class LotFragment extends MF.Fragment {
       return bldgclassLookup[this.bldgclass];
     }
 
-    @computed('cd')
-    get borocode() {
-      const borocd = this.cd;
-      return `${borocd}`.substring(0, 1);
+    @computed('borocode')
```
TylerMatteo, line 417: (just commenting here what we discussed IRL) - you should able to just use the `borocode` attribute created above, instead of needing a computed property for this.

.gitignore
```diff
@@ -18,6 +18,7 @@
 /npm-debug.log*
 /testem.log
 /yarn-error.log
+development.env
```
TylerMatteo, line 21: Why this change? I wasn't aware that Zola needed to kind of file.

TylerMatteo (changes requested):

.gitignore
TylerMatteo, line 21: Do you need this? I didn't think Zola used one of these files

app/models/map-features/lot.js
```diff
@@ -426,16 +422,16 @@ export default class LotFragment extends MF.Fragment {
     @computed('cd')
     get cdName() {
       const borocd = this.cd;
-      const boro = `${borocd}`.substring(0, 1);
+      const borocode = this.borocode; // eslint-disable-line prefer-destructuring
       const cd = parseInt(`${borocd}`.substring(1, 3), 10).toString();
-      return `${boroLookup[boro]} Community District ${cd}`;
+      return `${boroLookup[borocode]} Community District ${cd}`;
```
TylerMatteo, line 427: I thought we changed this to use the first digit of borocd for just this use case because of cases like Marble Hill being in a Bronx CD even though it's in Manhattan? And, side note, let's try to avoid turning of eslint rules if we can. Was this one giving you particular trouble for some reason?

app/templates/components/layer-record-views/tax-lot.hbs
```diff
@@ -105,7 +106,7 @@
             target="_blank"
             href="https://www1.nyc.gov/site/planning/zoning/districts-tools/special-purpose-districts-{{specialDistrict.boroName}}.page#{{specialDistrict.anchorName}}"
             onclick={{action this.captureOutboundLink "Special Purpose District" preventDefault=false}}
-            class="button"
+            class="button special-purpose-district-button"
```
TylerMatteo, line 109: This should be safe to remove

TylerMatteo (changes requested):

app/models/map-features/lot.js
horatiorosa: I recall we tried this to get the test to pass. After the test failure, I changed it back to use the `borocode` column. The tests passed with this  but according to CD profiles, Marble Hill, Manhattan is part of the Bronx CD8 🥴   
I changed the const to `cdborocode` to more accurately describe what it is, `borocode` didn't seem accurate to me since MH is in Manhattan.

.gitignore
horatiorosa: I cannot recall why I had that. It's gone!

PR #1095, opened 2023-01-31
12132 Add link to ZAP Search to the ULURP IDs

dhochbaum-dcp:
When viewing a Zoning Map Amendment, the ULURP number links to ZAP Search page for appropriate project.

If the API does not return exactly 1 result, or the network request fails, the text will remain as it was previously.

Closes [AB#12132](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/12132)


horatiorosa (approved):
LGTM

PR #1134, opened 2023-06-29
Replace sass with sass-embedded and update ember-cli-build to use it

TylerMatteo:
### Summary

Addresses [AB#13946](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/13946)

This PR switches out the `sass` package for `sass-embedded`. I came across this while doing Ember 4 upgrades via [this GH Issue](https://github.com/adopted-ember-addons/ember-cli-sass/issues/211). On my machine, this is cutting the SASS portion of the build time down from ~24 seconds to ~3 seconds.

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -2,6 +2,7 @@
 
 const EmberApp = require('ember-cli/lib/broccoli/ember-app');
 const babelPlugin = require('ember-auto-import/babel-plugin');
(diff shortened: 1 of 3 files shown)


horatiorosa (approved):
good find on the interwebs.
🏆

PR #1135, opened 2023-06-29
Update links to regular and historical zoning maps to point to OTI media server

TylerMatteo:
## Summary

This PR updates the links to regular and historical zoning maps that users see when they click on a tax lot to point to the files on the media server that MPE recently migrated them to.


horatiorosa (approved):
🚀

PR #1139, opened 2023-07-18
14192 Add Matomo to ZoLa

dhochbaum-dcp:
This PR adds our self-hosted Matomo Tag Manager to ZoLa

Closes [AB#14192](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/14192)

### DIFF
--- a/app/components/bookmarks/types/-default.js
+++ b/app/components/bookmarks/types/-default.js
@@ -30,10 +30,10 @@ export default class DefaultBookmark extends Component {
     });
 
     // GA
-    this.get('metrics').trackEvent('GoogleAnalytics', {
-      eventCategory: 'Download',
-      eventAction: `Downloaded Bookmark as ${format}`,
-      eventLabel: 'Export',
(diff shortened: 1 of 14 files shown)


app/components/bookmarks/types/-default.js
```diff
@@ -20,6 +20,11 @@ export default class DefaultBookmark extends Component {
       event_category: 'Bookmark',
       event_action: 'Deleted Bookmark',
     });
+    this.get('metrics').trackEvent('MatomoTagManager', {
```
TylerMatteo, line 23: Is there a reason you left the `gtag` call above this instead of deleting it like you did with the `GoogleAnalytics` metrics call elsewhere?

dhochbaum-dcp: gtag is GA4, GoogleAnalytics is universal analytics, which is no longer in use.  I left in instances of GA4, but replaced the GoogleAnalytics instances with Matomo.

PR #1145, opened 2023-08-09
Added Police Precinct Info

dhochbaum-dcp:
THE FEATURE _NO ONE_ ASKED FOR!!!

### DIFF
--- a/app/models/map-features/lot.js
+++ b/app/models/map-features/lot.js
@@ -312,6 +312,703 @@ const landuseLookup = {
   11: 'Vacant Land',
 };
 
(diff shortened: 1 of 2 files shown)


app/datasets/nypd_precincts.js
```diff
@@ -0,0 +1,695 @@
+export const nypdPrecinctLookup = [
```
TylerMatteo, line 1: This is better but my issue is that this is a front end and data doesn't go in front ends (or at least, it's not supposed to). We'll cross this bridge if we decide this is worth releasing.

PR #1157, opened 2023-11-22
1155 Update ZOLA DTM link

### LINKED ISSUE #1155: Investigate and resolve Zola DTM link issues
Several users have reported issues attempting to use the Digital Tax Map Link when viewing Tax Lots in Zola. They can click the link but get sent to a 404 or error page.

Image

What we know so far:
* The issue seems to be at least somewhat inconsistent to reproduce. When testing, be sure to test different browsers and different networks (citynet, office guest network, third party)
* The issue seems to be happening only when using Chrome, or at least reproduces most consistently when using Chrome
* The URL we link to (`maps.nyc.gov`) appears to be hosted with http, not https. If that is the root cause of the issue, we may need to reach out to OTI and/or Department of Finance (DOF)
* The application at `maps.nyc.gov` seems to also be reachable at `gis.nyc.gov`. This URL is also hosted with http, so it may not be useful, but it does exist.
* Matt Croswell reached out to a contact at DOF about this issue in mid-October, which is where some of this info originates. I'll forward that thread to the AE DL (subject line "DTM Map application") so whoever picks this up has it.

Next steps for us:
* Update the link in ZoLa to use `http://gis.nyc.gov/taxmap/map.htm`.
* Verify link to DTM still works in ZoLa staging environment
* Have DCP stakeholder verify the new link is working as well

If we still have issues after updating the link, the issue may be that our users are using browsers that block traffic to sites hosted on http (instead of https). If that is the case, we will have to reach out to DOF/OTI for assistance.

horatiorosa:
## Description

This PR updates the digital tax map link from `http://maps.nyc.gov/taxmap/map.htm` to `http://gis.nyc.gov/taxmap/map.htm`
in an effort to avoid a 400 on redirect on Chrome browser when redirected to the DOF hosted tax maps

## Tickets
Closes #1155

### DIFF
--- a/app/components/layer-record-views/tax-lot.js
+++ b/app/components/layer-record-views/tax-lot.js
@@ -441,7 +441,7 @@ export default class TaxLotRecordComponent extends LayerRecordComponent {
   }
 
   get digitalTaxMapLink() {
-    return `http://maps.nyc.gov/taxmap/map.htm?searchType=BblSearch&featureTypeName=EVERY_BBL&featureName=${this.model.bbl}`;
+    return `http://gis.nyc.gov/taxmap/map.htm?searchType=BblSearch&featureTypeName=EVERY_BBL&featureName=${this.model.bbl}`;
   }
 
   get zoningMapLink() {
--- a/tests/integration/components/layer-record-views/tax-lot-test.js
+++ b/tests/integration/components/layer-record-views/tax-lot-test.js
@@ -83,7 +83,7 @@ module(
 
       assert.equal(
         find('[data-test-tax-map-link]').getAttribute('href'),
-        'http://maps.nyc.gov/taxmap/map.htm?searchType=BblSearch&featureTypeName=EVERY_BBL&featureName=1000477501'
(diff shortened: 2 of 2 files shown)


TylerMatteo (approved):
@horatiorosa  looks good. Because `develop` has the upgrade changes that are still in QA, please open another PR for this feature branch against `master`.

horatiorosa:
> @horatiorosa looks good. Because `develop` has the upgrade changes that are still in QA, please open another PR for this feature branch against `master`.

I think I would need to open new PR bc unless I have lost my mind, the `digitalTaxmapLink` function is in [/app/models/map-features/lot.js](https://github.com/NYCPlanning/labs-zola/blob/master/app/models/map-features/lot.js#L545)  on `master`, while on `develop` it's in [app/components/layer-record-views/tax-lot.js](https://github.com/NYCPlanning/labs-zola/blob/develop/app/components/layer-record-views/tax-lot.js#L443).

I am assuming some refactoring was done.

I will proceed and make a new PR on a fresh branch from master.

PR #1169, opened 2023-12-29
Add missing mixed special purpose districts

TylerMatteo:
This PR adds several missing mixed used special purpose districts to the hardcoded values in environment.js. This includes two new ones added to the public website on 12/29/2023 - MX-24 and MX-25.


horatiorosa (approved):
🚀

PR #1185, opened 2024-03-22
Add Bookmarking Sets of Layers

dhochbaum-dcp:
This PR allows users to bookmark the current layers, and access their bookmarked layers in the saved menu.

image

Completes #1192

### DIFF
--- a/app/controllers/bookmarks.js
+++ b/app/controllers/bookmarks.js
@@ -15,6 +15,17 @@ export default Controller.extend({
 
   editMode: false,
 
+  track(act) {
+    gtag('event', 'saved_layer_sets', {
+      event_category: 'Saved Layer Sets',
+      event_action: act,
+    });
+    this.metrics.trackEvent('MatomoTagManager', {
+      category: 'Saved Layer Sets',
+      action: act,
+      name: act,
+    });
+  },
   // because we must compute the record types based on multiple
   // promises, the model uses Promise.all
   // this gets us in trouble when we need to do
@@ -58,6 +69,7 @@ export default Controller.extend({
       window.localStorage['saved-layer-sets'] = JSON.stringify(
         this.savedLayerSets
       );
+      this.track('bookmarkCurrentLayerSet');
     },
 
     deleteBookmarkedLayerSettings(id) {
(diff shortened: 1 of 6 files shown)


app/controllers/bookmarks.js
```diff
@@ -26,5 +46,115 @@ export default Controller.extend({
...
+      ];
+      layerGroups.forEach((lg) => {
+        lg.visible = !!layerToLoad.visibleLayerGroups.includes(lg.id);
+        lg.layers.forEach((layer) => {
+          layer.visibility = !!layerToLoad.visibleLayers.includes(layer.id);
+        });
+      });
+
+      [
+        'selectedZoning',
```
TylerMatteo, line 146: Can you move this list of strings that is repeated at line 63 into a variable? I see that line 64 adds `layerGroups`, so you could append that extra one there and use the variable for the rest.

app/controllers/bookmarks.js
```diff
@@ -6,6 +7,25 @@ import { Promise } from 'rsvp';
 export default Controller.extend({
   mainMap: service(),
   metrics: service(),
+  router: service(),
+
+  savedLayerSets: window.localStorage['saved-layer-sets']
```
TylerMatteo, line 12: This might be more trouble than it's worth but did you consider making an Ember model for these bookmarkable layer sets, similar to how we have a `bookmark` model already? This approach isn't bad but using a model might be more Ember-y

TylerMatteo (changes requested):

app/controllers/bookmarks.js
dhochbaum-dcp: Done

app/controllers/bookmarks.js
dhochbaum-dcp: I considered it, but abandoned the idea after looking at the way the bookmarks were set up and finding the abstraction was a little much for me.

TylerMatteo: @dhochbaum-dcp Have you spent any more time on the approach I described here? If you have and are still hitting a wall, I won't hold up the PR over it.

TylerMatteo: Ok. I'll take another look at the PR as is and leave a review.

PR #1194, opened 2024-04-03
Allow Multiple Measurements

dhochbaum-dcp:
Currently, clicking the distance or area button clears any previously drawn distance/area.  With this PR, you are now able to draw multiple lines or shapes, and able to see the sum of the drawn distances/areas.  The X button continues to clear all drawn shapes and stored measurements.

image

Completes #1195

### DIFF
--- a/app/components/map-measurement-tools.js
+++ b/app/components/map-measurement-tools.js
@@ -137,13 +137,8 @@ export default class MapMeasurementToolsComponent extends Component {
     this.set('draw', draw);
     const drawMode = type === 'line' ? 'draw_line_string' : 'draw_polygon';
     const { mainMap } = this;
-    if (mainMap.get('drawMode')) {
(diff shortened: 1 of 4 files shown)


app/components/map-measurement-tools.js
```diff
@@ -56,11 +138,12 @@ export default class MapMeasurementToolsComponent extends Component {
     const drawMode = type === 'line' ? 'draw_line_string' : 'draw_polygon';
     const { mainMap } = this;
     if (mainMap.get('drawMode')) {
-      draw.deleteAll();
+      // the following have been commented out, as we no longer want to clear all previous drawings unless the user hits the x button
```
TylerMatteo, line 141: No need to commit commented out code. If we don't need it, delete it. If we have to bring it back, that's what git is for.

TylerMatteo (changes requested):
Just clean up the commented code. Otherwise lgtm

Issue #1196, opened 2024-04-04
Pop Up Modal


dhochbaum-dcp:
Add a pop-up modal to the site to display on page load.  It should have:
- The option to not show the modal again when the user closes the window, or when the user checks the checkbox and closes the window - which option tbd
- A slideshow of new features

jpiacentinidcp:
Love these. I do think we will need to drop the background colors, but otherwise they are solid.

We can update the gif of the sidebar hide arrow once that feature is working properly. 

Let's discuss a potential blog post or other social media blast with Frank next week. I like the idea of doing that

jpiacentinidcp:
Hi David - So with this modal I think all we need to do is (sadly) remove the bright colored background and also remove the page on the zoning text. That feature also should be removed from the "new features" tab. 

Otherwise I think everything is good to go. I'll reach out to comms and ask about sharing these with the public once it's all good to merge in

dhochbaum-dcp:
@jpiacentinidcp I've removed the colored backgrounds.  I never added the zoning text to the slideshow, so that is already complete.

<img width="594" alt="image" src="https://github.com/NYCPlanning/labs-zola/assets/61206501/1594912a-3edd-4e5b-b2c1-958800d6587f">
This placeholder text needs to be updated.  What are we doing about a blog post?  We'll need that link as well.

jpiacentinidcp:
Sounds good. I will confirm how we want to proceed with comms early next week. I don't think we will link to a blog post but rather have them post something on Twitter/X or LinkedIn with a link to ZoLa - and this pop up window has static text.

jpiacentinidcp:
@dhochbaum-dcp @TylerMatteo Spoke with Jackie and Tim in comms, they estimate they can have a social media post ready to share by Thursday to highlight all the ZoLa updates. They can hold that in the wings until we are ready to merge and push to prod.

Do we have a sense of the timeline for remaining work across all the features? Would be good to get them all into one environment and then maybe have half a day to test them together.

David, can you include a screenshot here of the updated modal text just to confirm?

dhochbaum-dcp:
<img width="620" alt="image" src="https://github.com/NYCPlanning/labs-zola/assets/61206501/473619d8-509c-4b7e-abcd-ea72a94c1d39">

TylerMatteo:
[...]

PR #1197, opened 2024-04-04
Pop Up Modal

dhochbaum-dcp:
This PR adds a pop-up modal to the site to display on page load.  It has:
- The option to not show the modal again when the user closes the window, or when the user checks the checkbox and closes the window - currently, the latter is used
  - If you have checked the box and closed the window, you can enter `window.localStorage.clear()` in the developer console and refresh to be able to see it again
- A slideshow of new features

Completes #1196


app/controllers/application.js
```diff
@@ -200,6 +200,22 @@ export default class ApplicationController extends Controller.extend(
...
 
+  @tracked
+  openModal = !window.localStorage.hideMessage;
+
+  @tracked
+  dontShowModalAgain = false;
+
+  @action toggleModal() {
+    this.openModal = !this.openModal;
+    // Uncomment this when done so that the message only shows the first time
```
TylerMatteo, line 211: Can we remove these 3 commented lines now? I'd rather they not land in develop

TylerMatteo (changes requested):

app/controllers/application.js
dhochbaum-dcp: I've removed them, although I don't really understand why.  They make it easy for any developer in the future to switch between showing the modal once, showing the modal until the user hits the checkbox, or always showing the modal.  What's the drawback?

TylerMatteo (approved):

PR #1213, opened 2024-11-08
Add record for East Tremont Corridor special district to environment.js

TylerMatteo:
This PR adds a record to environment.js for the new [Bronx East Tremont Corridor special district](https://www.nyc.gov/site/planning/zoning/districts-tools/special-purpose-districts-bronx.page#ETC).


horatiorosa (approved):
🚀

PR #1227, opened 2025-04-02
Release: Bugfix: incorrect zoning map path

TangoYankee:
# Release
feature branch -> master

## Changes
Add leading zero to maps labeled with single digit


horatiorosa (approved):
⭐

Issue #1288, opened 2025-10-05
Add Special Atlantic Avenue Mixed Use District special purpose district


TylerMatteo:
# Acceptance Criteria

- [ ] New data for Special Atlantic Avenue Mixed Use District is added to `specialDistrictCrosswalk` in `/config/environment.js`
- [ ] When new SPD is selected, user sees SPD title from Carto
- [ ] "Learn more" link for SPD links user to [correct page and header on website](https://www.nyc.gov/content/planning/pages/zoning/zoning-districts-guide/special-purpose-districts#AAM)

PR #1289, opened 2025-10-06
Add aam to environment.js

pratishta:
Closes https://github.com/NYCPlanning/labs-zola/issues/1288

Fixes the "Learn more" link for Special Atlantic Avenue Mixed Use District by adding it explicitly so that the url can be generated from the tag "AAM"


TylerMatteo:
@pratishta can you add a little more to the PR description? then you can unmark as draft.

TylerMatteo (approved):

